PR #92, opened 2025-02-08
Add POST APIs for licenses import

arasHi87:
### What type of PR is this?

feat

#40 

### What this PR does?

* Provide two POST APIs: one for cluster license import and another for node license import. This is the first iteration and will be refactored in the future.
* Create a `DevSecret` middleware to handle API communication. It will read the secret from the header and bypass token verification, only for development until we implement proper API-to-API communication verification. The secret bypass should **be removed** before the M1 release.
* The `hex_sdk` always returns code 0 for license imports, regardless of success or failure. Need to check with the COS developer.

### Test results (optional)

* For cluster license importing, send the request to the `dell13` server and ensure it delegates the request to `dell200`, confirming that the license has been imported on `dell200`.
image

* Swagger of the cluster license import
image
image

* For node license importing, ensure the license has beed imported on the host
image

* Swagger of the node license import
image
image

### DIFF
--- a/api/docs.go
+++ b/api/docs.go
@@ -1254,6 +1254,158 @@ const docTemplate = `{
                         }
                     }
                 }
+            },
+            "post": {
+                "operationId": "importClusterLicense",
+                "tags": [
+                    "Licenses"
+                ],
+                "summary": "Update licenses for the cluster",
+                "parameters": [
+                    {
+                        "in": "path",
+                        "name": "dataCenter",
+                        "required": true,
+                        "schema": {
+                            "type": "string"
+                        },
+                        "description": "The name of the data center to operate",
+                        "example": "example-data-center"
+                    },
+                    {
+                        "in": "path",
+                        "name": "node",
+                        "required": true,
+                        "schema": {
+                            "type": "string"
+                        },
+                        "description": "The name of the node to operate",
+                        "example": "example-node"
+                    }
+                ],
+                "requestBody": {
+                    "required": true,
+                    "content": {
+                        "multipart/form-data": {
+                            "schema": {
+                                "$ref": "#/components/schemas/PostLicenseRequest"
+                            }
+                        }
+                    }
+                },
+                "responses": {
+                    "200": {
+                        "description": "Update licenses successfully",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/PostLicenseResponse"
+                                }
+                            }
+                        }
+                    },
+                    "500": {
+                        "description": "Internal Server Error",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "type": "object",
+                                    "properties": {
+                                        "code": {
+                                            "type": "integer",
+                                            "example": 500
+                                        },
+                                        "msg": {
+                                            "type": "string",
+                                            "example": "failed to update licenses: internal server error"
+                                        },
+                                        "status": {
+                                            "type": "string",
+                                            "example": "internal server error"
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        },
+        "/api/v1/datacenters/{dataCenter}/nodes/{node}/licenses": {
+            "operationId": "importNodeLicense",
+            "post": {
+                "tags": [
+                    "Licenses"
+                ],
+                "summary": "Update licenses for specific node",
+                "parameters": [
+                    {
+                        "in": "path",
+                        "name": "dataCenter",
+                        "required": true,
+                        "schema": {
+                            "type": "string"
+                        },
+                        "description": "The name of the data center to operate",
+                        "example": "example-data-center"
+                    },
+                    {
+                        "in": "path",
+                        "name": "node",
+                        "required": true,
+                        "schema": {
+                            "type": "string"
+                        },
+                        "description": "The name of the node to operate",
+                        "example": "example-node"
+                    }
+                ],
+                "requestBody": {
+                    "required": true,
+                    "content": {
+                        "multipart/form-data": {
+                            "schema": {
+                                "$ref": "#/components/schemas/PostLicenseRequest"
+                            }
+                        }
+                    }
+                },
+                "responses": {
+                    "200": {
+                        "description": "Update licenses successfully",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/PostLicenseResponse"
+                                }
+                            }
+                        }
+                    },
+                    "500": {
+                        "description": "Internal Server Error",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "type": "object",
+                                    "properties": {
+                                        "code": {
+                                            "type": "integer",
+                                            "example": 500
+                                        },
+                                        "msg": {
+                                            "type": "string",
+                                            "example": "failed to update licenses: internal server error"
+                                        },
+                                        "status": {
+                                            "type": "string",
+                                            "example": "internal server error"
+                                        }
+                                    }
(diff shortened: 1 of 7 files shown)


jdjgya (changes requested):
Hi @arasHi87 , 

Thanks for helping on the license features.

Based on the api practice we've discussed with FE team, please 

- add operationId for new APIs

- use $ref to associate the response schema

- add required fields for object keys

- add multi-example for most of API response descs, so that reader can more easier to understand the possible cases from API (if needed)

internal/api/v1/licenses/handlers.go
```diff
@@ -17,6 +17,18 @@ var (
...
+		{
+			Version: api.V1,
+			Method:  http.MethodPost,
+			Path:    "/licenses",
+			Func:    importClusterLicense,
+		},
+		{
+			Version: api.V1,
+			Method:  http.MethodPost,
+			Path:    "/nodes/:node/licenses",
```
jdjgya, line 29: the path hierarchy seems like a bit inconsistent between 

- POST /licenses
- POST /nodes/:node/licenses

for now and based on how k8s treat the sub-resources(like metrics) for pod and node, I would like to see if they could under the /licenses 🤔 

or maybe you can keep as it is, let me refactor them in the following iteration which is also okay 👌 maybe we'll see other scenarios as time goes on in the M1

internal/runtime/router.go
```diff
@@ -59,8 +60,27 @@ func initReqInfo(c *gin.Context) {
...
+func verifyDevSecret() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		secret := c.GetHeader("secret")
+		if secret == "Dev@Cube" {
+			c.Set("isSecretValid", true)
+		}
+
+		c.Next()
+	}
+}
```
jdjgya, line 74: Thanks for making this for team during current developing! 👍

internal/api/v1/licenses/handlers.go
arasHi87: You mean `/licenses/nodes/:node`?

I considered that before but felt it was a little bit weird. It seems like more fundamental resources should come earlier in the path?

But if there's already an example, I can follow it. By the way, do you know where Kubernetes has similar examples? I tried searching the codebase before but didn’t find anything.

arasHi87:
> Hi @arasHi87 ,
> 
> Thanks for helping on the license features.
> 
> Based on the api practice we've discussed with FE team, please
> 
> * add operationId for new APIs
> * use $ref to associate the response schema
> * add required fields for object keys
> * add multi-example for most of API response descs, so that reader can more easier to understand the possible cases from API (if needed)

Refactored, please help to check it again, thanks.

internal/api/v1/licenses/delegation.go
```diff
@@ -0,0 +1,54 @@
...
+	resp, err := h.R().
+		SetFileReader("license", licenseFile.Filename, reader).
+		SetHeader("secret", "Dev@Cube").
+		Post(genUrl(node))
+	if resp.IsError() || err != nil {
+		log.Errorf(
+			"failed to send license to node %s: %d %s",
+			node.Hostname,
+			resp.StatusCode(),
+			string(resp.Body()),
```
jdjgya, line 48: glad to see we choose to use go-resty under the hood.

their wrapped functions look pretty comfortable and beautiful to use

arasHi87: yeah, I agree with it, it's more nature and powerful ... 🚀

internal/api/v1/licenses/handlers.go
jdjgya: > re: I considered that before but felt it was a little bit weird. It seems like more fundamental resources should come earlier in the path?

<br/>

it could be, but I suppose it might also depends on how we define the relationship for those resources.

here's my tmp thoughts currently 🤔 

- if you tend to put the nodes in the first layer, then I think you might consider to put it under the `nodes` dir? to let the node handlers to be placed together.

- we can postpone the final decision for the path shape, may be we will have other thoughts in the incoming tasks.


<br/>

> re: But if there's already an example, I can follow it. By the way, do you know where Kubernetes has similar examples? I tried searching the codebase before but didn’t find anything.

It's pretty simple to search from their doc or cloud vendor's doc

https://kubernetes.io/docs/tasks/debug/debug-cluster/resource-metrics-pipeline/

https://docs.okd.io/latest/rest_api/monitoring_apis/podmetrics-metrics-k8s-io-v1beta1.html

jdjgya: btw, k8s have an early repo to place the design doc https://github.com/kubernetes/design-proposals-archive/, but it's archived now, currently, they use the https://github.com/kubernetes/enhancements/ to track and discuss

not sure if some related design docs left there, so that we can get more concrete ideas from what they thought before.

api/docs.go
```diff
@@ -1254,6 +1254,178 @@ const docTemplate = `{
...
+                        "description": "The name of the node to operate",
+                        "example": "example-node"
+                    }
+                ],
+                "requestBody": {
+                    "required": true,
+                    "content": {
+                        "multipart/form-data": {
+                            "schema": {
+                                "type": "object",
```
jdjgya: sorry that I missed this one and should let you know earlier.

by the current practice we confirmed with team, 

if would be nice if the request schema can also be associated by `$ref`

api/docs.go
```diff
@@ -1254,6 +1254,178 @@ const docTemplate = `{
...
+                        },
+                        "description": "The name of the node to operate",
+                        "example": "example-node"
+                    }
+                ],
+                "requestBody": {
+                    "required": true,
+                    "content": {
+                        "multipart/form-data": {
+                            "schema": {
```
jdjgya, line 1367: same as this one https://github.com/bigstack-oss/cube-cos-api/pull/92/files#r1947936171

jdjgya (approved):
press approve first.

should be okay to merge after the remaining api-doc changes are updated(request schema associated by $ref) and rebase the branch (response schema looks good from your changes).

<br/>

The license and node pattern would be okay to leave it as current status 👌

internal/api/v1/licenses/handlers.go
arasHi87: Just thinking, whether it's `/nodes/:node/licenses` with the handler under `nodes` or `/licenses/nodes/:node`, both feel a bit weird to me personally.

Maybe we could merge the cluster and node license imports into a single endpoint and include the type (node or cluster) and node-specific parameters in the request body?

But I agree with you, we can keep it as is for now and wait for more examples before deciding how to converge it. Thanks for your advice and reference! :)))

jdjgya: it's worth discussion.

currently, I feel like it might be more close to the `/licenses/nodes/:node` because

<br/>

1). I feel like the license system is more close to a peer resource with cluster or node from COS's current design or the common cloud management.

<br/>

2). for example: if under your perspective, the most of PVC, Volume, or any of attach resources's related APIs should all under the PODs, VMs, Servers, but most of real cases are not like that if you see how k8s, openstack, or AWS treat this.

they're separated by the scenarios which they're going to provide, from it to identify which is the major entity

<br/>

3). I'll also try to figure out them, but I suppose the shape can't leave out some directions like

- who's the major entity(主體) of the case we're going to provide

- in case above, the major entity is License, but not the node,

- unless, designer design the flow like "License" > "Import License" > "Jump to Node page" > "Select Nodes to Import", then I will agree `/nodes/:node/licenses` is more close to the hierarchy we need, but meanwhile, this should be placed to the nodes rather licenses.

- to me, the "/nodes/:node/licenses" under the "license" handler is more awkward because the major entity is pretty confused whatever we try to describe the story from license side or node side.

Issue #49, opened 2025-01-29
[APIs Event] InfluxDB - Add continues query to do statistic pre-aggregation


jdjgya:
<br/>

### ▎Goal

<br/>

Store the event pre-aggregation result to InfluxDB, so that API will be able to fetch the different granularity of event statistic.

<br/>

The query duration is fixed on UI with the options below

- Last 14 days

- Last 7 days

- Last 24 hours(default)

- Last hour

<br/>

---

<br/>

### ▎Work Directions/Items (TDB with BE team)

<br/>

**1). system event - 1 hour aggregation measurement**

- time

- tag: eventId

- tag: severity

- tag: category

- field: number (int or float both are ok)

example:

```
> select * from testSystem1hr;
time                id        severity category  number
----                --        -------- --------  ------
1739713241024197254 SRV00001I critical service   10    
1739713241116557227 SRV00001I critical service   10    
1739713241128751796 SRV00001I info     service   10    
1739713241172672033 NET00001I critical net       20    
1739713241221256534 NET00001I critical net       20    
1739713241237652546 NET00001I info     net       20    
1739713241257592472 CEP00002W critical ceph      30    
1739713241285883263 CEP00002W critical ceph      30    
1739713242332652900 CEP00002W info     ceph      30    
```

<br/>

**2). host event - 1 hour aggregation measurement**

- time

- tag: eventId

- tag: host

- tag: category

- field: number (int or float both are ok)

example:

```
> select * from testSystem1hr;
time                id        host     category  number
----                --        -------- --------  ------
1739713241024197254 SRV00001I dell111  service   10    
1739713241116557227 SRV00001I dell112  service   10    
1739713241128751796 SRV00001I dell113  service   10    
1739713241172672033 NET00001I dell111  net       20    
1739713241221256534 NET00001I dell111  net       20    
1739713241237652546 NET00001I dell112  net       20    
1739713241257592472 CEP00002W dell111  ceph      30    
1739713241285883263 CEP00002W dell113  ceph      30    
1739713242332652900 CEP00002W dell112  ceph      30    
```

<br/>

**3). instance event - 1 hour aggregation measurement**

- time

- tag: eventId

- tag: instance

- tag: category

- field: number (int or float both are ok)

example:

```
> select * from testSystem1hr;
time                id        instance category  number
----                --        -------- --------  ------
1739713241024197254 SRV00001I vm-111   service   10    
1739713241116557227 SRV00001I vm-111   service   10    
[...]

Issue #24, opened 2025-01-13
[APIs Events] List - GET API to list events


jdjgya:
<br/>

▎GET /api/v1/datacenters/{data-center}/events?pages=1&perPage=3

2XX response example:

```
{
    "code": 200,
    "status": "ok",
    "msg": "fetch events successfully",
    "data": [
        {
            "type": "info",
            "id": "COMPUTE000031",
            "description": "instance 'vm-0001' is created on 'compute-0001'",
            "host": "compute-0001",
            "category": "compute",
            "service": "nova",
            "metadata": {
                "instance": "vm-0001",
                "host": "compute-0001"
            },
            "time": "yyyy/mm/dd hh:mm"
        },
        {
            "type": "error",
            "id": "NET000032",
            "description": "instance 'vm-0001' can't be deleted on 'compute-0001'",
            "host": "compute-0001",
            "category": "compute",
            "service": "nova",
            "metadata": {
                "instance": "vm-0001",
                "host": "compute-0001"
            },
            "time": "yyyy/mm/dd hh:mm"
        },
        {
            "type": "error",
            "id": "NET000032",
            "description": "instance 'vm-0001' can't be deleted on 'compute-0001'",
            "host": "compute-0001",
            "category": "compute",
            "service": "nova",
            "metadata": {
                "instance": "vm-0001",
                "host": "compute-0001"
            },
            "time": "yyyy/mm/dd hh:mm"
        }
    ]
}

```

<br/>

code >= 400 response example:

```
{
    "code": 400,
    "status": "bad request",
    "msg": "query parameter 'offset' is invalid"
}
```

PR #100, opened 2025-02-16
Add a GET API to fetch event statistic

jdjgya:
### What type of PR is this?

Feat and refactor

<br/>

Major: #24 #49 

<br/>

### What this PR does?

Major: as discussed with team in the kickoff of event page,

- add a GET API to fetch event statistic (ranking)

- enrich the event listing API to support query by event id

- based on 2 items above to support cascade event query from the event statistic (hope can bring some convenience to FE for saving their effort to concat query url)

- add corresponding changes in the swagger doc

<br/>

📔 notice:

- originally, I was implementing this by fake data, but during playing some real data in our current system, I realized the aggregation speed of influxdb is quick fast than I imagined even for 90 days search with huge result.

- moreover, the largest time range that team made in M1 is only 14 days, so I still do real time calculation in the influxdb currently, rather than waiting for COS CQ/Flux Task (will check with team again)

<br/>

### Test results (optional)

1). make sure the API doc has been updated accordingly

https://github.com/user-attachments/assets/95865984-5272-4283-8f75-b8ca98cdeaf4

<br>

2). make sure the mentioned apis can work properly

https://github.com/user-attachments/assets/a3346234-d3f6-45d7-b903-0311788267d6

### DIFF
--- a/api/docs.go
+++ b/api/docs.go
@@ -600,6 +600,16 @@ const docTemplate = `{
                     },
                     {
                         "in": "query",
+                        "name": "id",
+                        "required": false,
+                        "schema": {
+                            "type": "string"
+                        },
+                        "description": "The id of the event to query.",
+                        "example": "NET00003I"
+                    },
+                    {
+                        "in": "query",
                         "name": "start",
                         "required": false,
                         "schema": {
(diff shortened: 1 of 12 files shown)


jdjgya:
update:

self review finished, no any local settings left in the code.

might have adjustments afterward when FE start the integration.

PR #101, opened 2025-02-17
Implement CRUD operition for the setting APIs

arasHi87:
### What type of PR is this?

feat

Major: #50, #51, #52, #53 

### What this PR does?

* The first iteration of the Settings API implements CRUD operations for managing email sender info, email recipients, and Slack webhooks in MongoDB.
* Will consider merging the CRUD APIs for settings into one after discussing the trigger feature to ensure no additional requests are needed.

### Test results (optional)

All the endpoints have been tested in the local environment to ensure functionality.

#### CRUD of the email senders

* POST 
image
image

* GET
image
image

* PUT
image
image
image

* DELETE
image
image

#### CRUD of the email recipients

* POST 
image
image

* GET
image
image

* PUT
image
image

* DELETE
image
image

#### CRUD of the slack webhook

* POST
image
image

* GET
image
image

* PUT
image
image

* DELETE
image
image


arasHi87:
Just realized I forgot to add a settings API to retrieve all settings in one request. I'll add it later, but I think everything else is working fine.

internal/api/v1/settings/record.go
```diff
@@ -0,0 +1,194 @@
...
+	h := cubeMongo.GetGlobalHelper()
+	senders := []v1.EmailSender{}
+	cursor, err := h.GetQueryCursor(v1.SettingsDB(), v1.EmailSenderCollection(), bson.M{"deleted": bson.M{"$ne": true}})
+	if err != nil {
+		log.Errorf("failed to get cursor for email sender (%s)", err.Error())
+		return senders, err
+	}
+	defer cursor.Close(context.Background())
+
+	for cursor.Next(context.Background()) {
```
jdjgya: for avoiding the potential hanging,

let's try to use the `context.WithTimeout()` in the cursor.Next() 👍

internal/api/v1/settings/record.go
```diff
@@ -0,0 +1,194 @@
...
+
+func getEmailSenderRecords() ([]v1.EmailSender, error) {
+	h := cubeMongo.GetGlobalHelper()
+	senders := []v1.EmailSender{}
+	cursor, err := h.GetQueryCursor(v1.SettingsDB(), v1.EmailSenderCollection(), bson.M{"deleted": bson.M{"$ne": true}})
+	if err != nil {
+		log.Errorf("failed to get cursor for email sender (%s)", err.Error())
+		return senders, err
+	}
+	defer cursor.Close(context.Background())
```
jdjgya: for avoiding the potential hanging,

let's try to use the context.WithTimeout() in the cursor.Close() 👍

internal/api/v1/settings/record.go
```diff
@@ -0,0 +1,194 @@
...
+
+	for cursor.Next(context.Background()) {
+		sender := v1.EmailSender{}
+		if err := cursor.Decode(&sender); err != nil {
+			log.Errorf("failed to decode email sender record (%s)", err.Error())
+			continue
+		}
+		senders = append(senders, sender)
+	}
```
jdjgya, line 73: generally, we will do `if cursor.Err() != nil` to see if there's any error happens.

let's add it also 👍

internal/definition/v1/settings.go
```diff
@@ -0,0 +1,49 @@
...
+	ID      string `json:"id" bson:"id"`
+	Deleted bool   `json:"-" bson:"deleted"`
+
+	To   []string `json:"to" bson:"to"`
+	Note string   `json:"note,omitempty" bson:"note,omitempty"`
+}
+
+type SlackWebhook struct {
+	ID     string `json:"id" bson:"id"`
+	Delete bool   `json:"-" bson:"deleted"`
```
jdjgya: might be `Deleted` but not `Delete`?

jdjgya:
Hi @arasHi87 , Thanks a lot for helping on this!!

I just done the 1st round quick review by major checking the items below

- API pattern (the relationship between major entity and sub resources)

- DB operation details

  - let's try to use `ctx, canel := context.WithTimeout(wait.CtxSeconds(10))` for all the cursor.Next(ctx) and cursor.Close(ctx) in this iteration, for trying to avoid the infinity hanging from the uncontrollable/unstable env.

  - also let's try to check the `cursor.Err()` which is also an important part for using cursor

- Naming convention

- Error handling

- Function structure

internal/api/v1/settings/record.go
```diff
@@ -0,0 +1,194 @@
...
+	if err := h.UpdateOne(
+		v1.SettingsDB(),
+		v1.EmailSenderCollection(),
+		bson.M{},
+		bson.M{"$set": bson.M{"deleted": true}},
+	); err != nil {
+		log.Errorf("failed to delete email sender record (%s)", err.Error())
+		return err
+	}
+	return nil
```
jdjgya: I suppose the such kind of function struct can be simplified to 

```sh
func deleteEmailSenderRecord() error {
	h := cubeMongo.GetGlobalHelper()
        return h.UpdateOne(
		v1.SettingsDB(),
		v1.EmailSenderCollection(),
		bson.M{},
		bson.M{"$set": bson.M{"deleted": true}},
        )
}
```

unless, you really need the logs here, but I saw you will also capture the error outside the `deleteEmailSenderRecord()`, so maybe can consider to simplify this part and other parts which wrote like this 👍

internal/definition/v1/settings.go
```diff
@@ -0,0 +1,49 @@
+package v1
+
+const (
+	Settings = "settings"
+)
+
+type EmailSender struct {
+	Deleted bool `json:"-" bson:"deleted"`
```
jdjgya, line 17: not sure if we've to make a decision here,

but is it possible to use `Status` rather than deleted?

for example: status: "deleted"

because the status is more general to indicate a real situation of a resource, but you may have some thoughts for this, may I know your idea?

jdjgya:
Hi @arasHi87 , 

During quick review in your API Doc,

I realized some of response example seems like only has `msg` in the payload

```
                    "201": {
                        "description": "Email sender created successfully",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "type": "object",
                                    "properties": {
                                        "msg": {
                                            "type": "string",
                                            "example": "email sender created successfully"
                                        }
                                    }
                                }
                            }
                        }
```

but according the the newest design between FE and BE, it needs other info like `status` and `code`.

the above situation seems happens in many of interfaces you just wrote in the API doc, please help to check and add those info back, Thanks! 🙏

internal/api/v1/settings/record.go
```diff
@@ -0,0 +1,181 @@
...
+		v1.SlackWebhookCollection(),
+		webhook,
+	)
+}
+
+func getSlackWebhookRecords() ([]v1.SlackWebhook, error) {
+	h := cubeMongo.GetGlobalHelper()
+	webhooks := []v1.SlackWebhook{}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
```
jdjgya: 1 second might be too short?

btw, there's a `wait` lib in the dependency-go, you can use it with the context.WithTimeout from `context.WithTimeout(context.Background(), 1*time.Second)` to `context.WithTimeout(wait.CtxSeconds(1))`

internal/api/v1/settings/record.go
```diff
@@ -0,0 +1,181 @@
...
+	defer cancel()
+
+	cursor, err := h.GetQueryCursor(v1.SettingsDB(), v1.EmailSenderCollection(), bson.M{"deleted": bson.M{"$ne": true}})
+	if err != nil {
+		log.Errorf("failed to get cursor for email sender (%s)", err.Error())
+		return senders, err
+	}
+	defer cursor.Close(ctx)
+
+	for cursor.Next(ctx) {
```
jdjgya: generally, I remember ctx is better to not share in different function call which might have different lifecycle.

unless is using in the same scope of process chain

internal/definition/v1/settings.go
arasHi87: I thought about this too.

Used `deleted` because I see this as a configuration with a dual status, either existing or deleted, just like we discussed for K8s before.

But if we want to use `status: deleted` to manage it, that's fine with me too. Maybe in the future, we’ll want to extend it with other statuses like `using` or something similar.

That said, what about existing configurations? Should we leave the status field blank or set it to something like `created`? Don't have any good idea of the wording.

arasHi87: update.

Looks like K8s uses `DeletionTimestamp` to determine if a resource has been deleted. Maybe we can consider using `deleted_at`, `updated_at`, and `created_at` instead. That way, we can also use them for audit logs.

May need more discussion.

jdjgya (approved):

PR #117, opened 2025-02-27
CI: Build RPM, Install RPM to Staging, and Adjust Configs to Deploy to CubeCOS

Eandalf-Bigstack:
### What type of PR is this?

ci

### What this PR does?

- Add rpm spec file
- Set a build script to build rpm
- Add README.md about installing cube-cos-api to staging using rpm
- Adjust configs to deploy to CubeCOS
- This PR depends on https://github.com/bigstack-oss/bigstack-dependency-go/pull/3 , would need to wait for the mentioned PR to be merged and adjust `go.mod` in this PR


jdjgya:
Thanks a lot @Eandalf-Bigstack 🙏

I've taken a look and probably understand your thoughts.

would you mind let me adjust your code? I'll follow the major directions you want, just org it a bit, some of features will be extended from dependency-go

jdjgya:
Hi @Eandalf-Bigstack , 

I've fixed the issue from your original code base and simplify some of logic or code struct.

btw, before refactor, running `./rpm.sh` on my MAC will get the error below

Screenshot 2025-02-28 at 3 35 51 AM

<br/>

after refactor, I still get this, so might be not impacted by refactor.

during investigating, it seems like related to the macro about `BuildRequires: systemd`,

but I saw you already placed it in the spec file, so not sure if it's my env problem.

I guess that you run the rpm process in the amd64(x86_64) system, rather than arm64, right?

if so, then I think I can skip the error from mac

Eandalf-Bigstack:
Thank you for helping me refactoring the code! And, adding the script to CI is great.

> after refactor, I still get this, so might be not impacted by refactor.

As for this issue, I reproduced this on mac also.

I used `brew install rpm` to get the build tools ready on mac. I added `/` in front of the `%{_unitdir} and got the build go through without complaining errors. So, this might be the difference caused by the rpm build tool on mac (from brew) and the build tool on CentOS (from dnf/rpm).

Adding this change would help. However, we are building the rpm package targeting `amd64` architecture. I just tried and the build tool refused to cross compile an `amd64` rpm package on an `arm64` mac. So, we would need it to be run on an `x86_64` machine. Previously, I was building the rpm package using CentOS9 on `x86_64`. Since we could not build the target rpm package on mac, I would like to leave the issue as it and add the requirement of the machine architecture for building the rpm in `README.md`.

By the way, I forgot to add the build tool installation in `README.md`. I would add this part in the coming commit.

jdjgya:
> I used `brew install rpm` to get the build tools ready on mac. I added `/` in front of the `%{_unitdir} and got the build go through without complaining errors. So, this might be the difference caused by the rpm build tool on mac (from brew) and the build tool on CentOS (from dnf/rpm).
> 

Got it - Thanks for confirming 🙏 

<br/>

> Adding this change would help. However, we are building the rpm package targeting `amd64` architecture. I just tried and the build tool refused to cross compile an `amd64` rpm package on an `arm64` mac. So, we would need it to be run on an `x86_64` machine. Previously, I was building the rpm package using CentOS9 on `x86_64`. Since we could not build the target rpm package on mac, I would like to leave the issue as it and add the requirement of the machine architecture for building the rpm in `README.md`.

okay - Got it. I think it should be fine 👌 

<br/>


> By the way, I forgot to add the build tool installation in `README.md`. I would add this part in the coming commit.

no worries - Today is national holiday btw, remember to have some rest

jdjgya (approved):

PR #133, opened 2025-03-11
feat(login): adjust saml sp to connect ui

Eandalf-Bigstack:
### What type of PR is this?

- feature

- Connect the login process between API and UI

### What this PR does?

- Currently, the login process would encounter two obstacles, when accessing a resource with the authentication requirement, the redirect issued by the service provider (samlsp) would be blocked by the browsers due to CORS. The second obstacle is that after finishing login, the redirect URL would be API's resource URI, which would redirect the browser to see JSON outputs.
- To tackle the first issue, I tried upgrading Keycloak to have the option to set "Web Origins" for CORS HTTP headers. However, keycloak does not support "Web Origins" options for SAML clients, only support the option for OIDC clients.
  - https://github.com/keycloak/keycloak/issues/30759
  - Hence, I provided our own implementation to do SAML auth (function saml.DoSamlAuth). The implementation mostly came from the one in `saml/samlsp`, but it would send the redirect URL for Keycloak as payload `msg` back in the 401 response, instead of either directly issuing 302 redirect or responding with the login form.
- To tackle the second issue, I added our own implementation to issue the redirect in function saml.ServeAcs. The implementation is also mostly identical to the one in `saml/samlsp`. However, we only supply the default path `/`, instead of the path pointing to an API resource, leading to JSON.

### DIFF
--- a/configs/cube-cos-api.yaml.template
+++ b/configs/cube-cos-api.yaml.template
@@ -13,10 +13,9 @@ spec:
       system: /etc/settings.sys
     logoutRedirect: "/api/v1/datacenters"
     keycloak:
-      # host: https://10.32.10.180:10443/auth
       host:
         scheme: "https"
-        ip: "10.32.10.180"
+        ip: ""
         port: 10443
         path: "auth"
         tlsInsecureSkipVerify: true
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,6 @@ require (
 	github.com/micro/plugins/v5/wrapper/ratelimiter/uber v1.0.2
 	github.com/mohae/deepcopy v0.0.0-20170929034955-c48cc78d4826
 	github.com/shirou/gopsutil/v4 v4.25.1
-	github.com/slack-go/slack v0.16.0
 	github.com/swaggo/files v1.0.1
 	github.com/swaggo/gin-swagger v1.6.0
 	github.com/swaggo/swag v1.16.3
@@ -126,6 +125,7 @@ require (
 	github.com/russellhaering/goxmldsig v1.3.0 // indirect
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
 	github.com/segmentio/ksuid v1.0.4 // indirect
+	github.com/slack-go/slack v0.16.0 // indirect
 	github.com/tklauser/go-sysconf v0.3.12 // indirect
 	github.com/tklauser/numcpus v0.6.1 // indirect
 	github.com/twitchyliquid64/golang-asm v0.15.1 // indirect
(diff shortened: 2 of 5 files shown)


configs/cube-cos-api.yaml.template
```diff
@@ -13,10 +13,9 @@ spec:
       system: /etc/settings.sys
     logoutRedirect: "/api/v1/datacenters"
     keycloak:
-      # host: https://10.32.10.180:10443/auth
       host:
         scheme: "https"
-        ip: "10.32.10.180"
```
jdjgya, line 19: oooops Thanks a lot for catching this.

Sorry that I committed this by by accident o_q

jdjgya:
Hi @Eandalf-Bigstack , before adjust the code, 

I've tried your implementations on my 10.32.10.180, but looks like it will be failed by the error below

Screenshot 2025-03-13 at 10 09 52 PM

<br/>

I've also tried on your 2 testing env because I guess you might have the similar env on them,

- https://10.32.34.1:4443/api/v1/datacenters

- https://10.32.35.1:4443/api/v1/datacenters

but they all get the same error above.

<br/>

not sure if there's something wrong? or needs extra settings for COS first?

Eandalf-Bigstack:
> I've tried your implementations on my 10.32.10.180, but looks like it will be failed by the error below

This is the intentional behavior. Due to CORS, api cannot directly issue redirect (302) to the browser, which was the previous behavior. So, I let SAML service provider send the redirect URL back to browser in the unauthorized (401) payload, the `msg` field, then the https://github.com/bigstack-oss/cube-cos-ui/blob/40a84c1f58abf480d074246efc405d459a6e95a2/packages/cube-frontend-web-app/src/api/samlAuthErrorInterceptor.ts#L19-L21 part of UI would handle the redirect. Hence, this approach avoid the CORS.

For the 401 response without UI, we could copy paste the URL supplied in `msg` field of the 401 response to a new tab, and fill out the username and password in Keycloak, then we could come back to the api page, then reload, then we should be authorized.

jdjgya:
> https://github.com/bigstack-oss/cube-cos-ui/blob/40a84c1f58abf480d074246efc405d459a6e95a2/packages/cube-frontend-web-app/src/api/samlAuthErrorInterceptor.ts#L19-L21

ahhh okay - gotcha!

Thank you for the detailed explanation, now I understand the flow 🙏 

and Yes - copy paste the url from the 401 page can see the login page again 👍

jdjgya:
Hi @Eandalf-Bigstack ,

<br/>

Thanks a lot for your implementation, it's amazing for trying to resolve  CORS problem, learned!

<br/>

and as discussed yesterday, I've adjusted the code struct and the flow a bit by the `guard clause` (https://en.wikipedia.org/wiki/Guard_(computer_science)), I've tested the functionality in my 10.180 which works well also.

<br/>

for your reference,

Although, guard clause is not suitable for all the cases, but when I encounter some nested flows, I will still try to use it and extracting some of lines to an individual func to flatten the code.

It's an easy and interesting concept to play, and in most of the time, it can make the high level logic to be more clear in the code 🙏 

Screenshot 2025-03-14 at 12 20 57 AM

Screenshot 2025-03-14 at 12 21 06 AM

jdjgya:
have tested the features, LGTM, let's merge it first 😎 

Screenshot 2025-03-14 at 12 15 11 AM

Screenshot 2025-03-14 at 12 15 23 AM

jdjgya (approved):

PR #163, opened 2025-03-31
refactor(ci): use find to make mv more robust in rpm spec

Eandalf-Bigstack:
### What type of PR is this?

- refactor

### What this PR does?

- Use find to make mv more robust in rpm spec


jdjgya (approved):
LGTM!

PR #175, opened 2025-04-08
feat(ha): add api health check endpoint for the load balancer haproxy

Eandalf-Bigstack:
### What type of PR is this?

- feature

### What this PR does?

- Add a health check endpoint for the load balancer HAProxy

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ toolchain go1.24.0
 
 require (
 	github.com/Nerzal/gocloak/v13 v13.9.0
+	github.com/aws/aws-sdk-go-v2/service/s3 v1.78.2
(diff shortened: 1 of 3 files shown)


jdjgya:
Hi @Eandalf-Bigstack , 

Thanks so much for raising this for us 👍 !!

but sorry that I just adjusted it a bit to be `/live` and `livenessCheck()`. 

because I would like to make the api health check could be more detailed and specific in different directions in the future like how k8s does https://kubernetes.io/docs/reference/using-api/health-checks/ 😃

jdjgya (approved):

Eandalf-Bigstack:
Got it, I would also adjust the HAProxy liveness probe endpoint setting to `/live` in cubecos.

PR #582, opened 2026-06-04
fix(tunings): select the frontend storage network interface only

Eandalf-Bigstack:
### What type of PR is this?

- fix

- When both frontend and backend storage network are provided to a node, `cube-cos-api` would fail to start.

### What this PR does?

- Parse the storage network tuning correctly if both frontend and backend storage network are provided

### Test results (optional)

<br/>


arasHi87 (approved):
LGTM

PR #585, opened 2026-06-08
Chore/build with docker due to nvml

steven-chiu-bigstack:
### What type of PR is this?

Chore

None

### What this PR does?

Build GO binary in a Docker container because it fails to build on MacOS after NVML dependency is added to the project.

### Notes

Build process stays the same (run `task build` or `task buildLocalBinary`).

### Test results (optional)

None


steven-chiu-bigstack:
Replaced by 586

PR #593, opened 2026-06-23
feat: sync list gpu cards with new sdk usage

steven-chiu-bigstack:
### What type of PR is this?

Feat

Related to https://github.com/bigstack-oss/cubecos/issues/859

### What this PR does?

Sync list GPU cards API implementation with [the new SDK usage](https://github.com/bigstack-oss/cubecos/pull/934).

### Test results (optional)

<br/>

### DIFF
--- a/docs/developing/README.md
+++ b/docs/developing/README.md
@@ -102,7 +102,7 @@ So we don't need to manually update it every time.
 
 ### Prerequisites
 
-- [yq](https://github.com/mikefarah/yq) (for YAML → JSON conversion)
+- [yq](https://github.com/mikefarah/yq) >= 4.12.0 (for YAML → JSON conversion)
 
 ### Architecture
 
--- a/internal/apis/v1/handlers/nodes/gpu.go
+++ b/internal/apis/v1/handlers/nodes/gpu.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
-	"strconv"
 	"strings"
 
 	"github.com/NVIDIA/go-nvml/pkg/nvml"
@@ -104,7 +103,8 @@ func (h *helper) listLocalGpuCards() ([]gpu.GpuCard, error) {
 			return nil, err
 		}
 
-		vgpuProfiles, hexProfilesMap := listVgpuProfiles(device, hexGpu)
+		hexProfilesMap, hexProfileCollection := cubecos.GetNodeVgpuProfilesMap(hexGpu.PciAddress)
+
 		attachedInstances := listAttachedInstances(listAttachedInstancesOpts{
 			Device:                   device,
 			DeviceUUID:               uuid,
@@ -116,22 +116,20 @@ func (h *helper) listLocalGpuCards() ([]gpu.GpuCard, error) {
 			HexProfilesMap:           hexProfilesMap,
 		})
 
-		if vgpuProfiles != nil && attachedInstances != nil {
-			updateVgpuProfilesRemaining(*vgpuProfiles, *attachedInstances)
-		}
+		profileCollection := toProfileCollection(hexProfileCollection, attachedInstances)
 
 		gpuCards = append(gpuCards, gpu.GpuCard{
 			Id:                   hexGpu.Id,
 			Name:                 hexGpu.Name,
 			ResourceType:         hexGpu.Type,
 			SupportResourceTypes: hexGpu.SupportTypes,
-			Vram: &gpu.VramInfo{
+			Vram: gpu.VramInfo{
 				AllocatedMiB:       memoryUsedMiB,
 				TotalMiB:           memoryTotalMiB,
-				UtilizationPercent: float64(memoryUtilizationPercent),
+				UtilizationPercent: memoryUtilizationPercent,
 			},
-			Gpu: &gpu.GpuInfo{
-				UtilizationPercent: float64(gpuUtilizationPercent),
+			Gpu: gpu.GpuInfo{
+				UtilizationPercent: gpuUtilizationPercent,
 			},
 			PciAddress: pciAddress,
 			Status: gpu.GpuStatusInfo{
@@ -139,9 +137,8 @@ func (h *helper) listLocalGpuCards() ([]gpu.GpuCard, error) {
 				IsProcessing: false,
 			},
 			AllocationSummary: hexGpu.Allocation,
-			VramLimitMiB:      memoryTotalMiB,
 			ProfileCountLimit: hexGpu.ProfileCountLimit,
-			Profiles:          vgpuProfiles,
+			Profiles:          profileCollection,
 			AttachedInstances: attachedInstances,
 		})
 	}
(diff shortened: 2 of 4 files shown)


docs/developing/README.md
```diff
@@ -102,7 +102,7 @@ So we don't need to manually update it every time.
 
 ### Prerequisites
 
-- [yq](https://github.com/mikefarah/yq) (for YAML → JSON conversion)
+- [yq](https://github.com/mikefarah/yq) >= 4.12.0 (for YAML → JSON conversion)
```
steven-chiu-bigstack, line 105: I followed the [installation guide](https://github.com/mikefarah/yq#install) from the yq repository and found that `-o` was not supported in 4.2.0

Image

After some testing, I think the `-o` flag was added in 4.12.0.

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -420,20 +375,77 @@ func buildInstanceLinks(vmId string) gpu.InstanceLinks {
...
+		remainingMap[profile.Id] = profile.Count
+		profileIdMap[*profile.Alias] = profile.Id
+	}
+
+	for _, instance := range *attachedInstances {
+		if instance.ProfileAlias == nil || len(*instance.ProfileAlias) == 0 {
+			continue
+		}
+
+		profileId := profileIdMap[*instance.ProfileAlias]
```
raven-pan: Maybe we should use ok to check is profile exist:

```go
profileId, ok := profileIdMap[*instance.ProfileAlias]
if ok {
  remainingMap[profileId] = max(remainingMap[profileId]-1, 0)
}
```

internal/cubecos/nodes.go
```diff
@@ -421,41 +421,51 @@ func listNodeGpus(nodeName string) ([]gpu.GpuFromHex, error) {
...
 
 	out, err := exec.CommandContext(ctx, "hex_sdk", "gpu_vgpu_profile_list", "-gpuId", gpuId).CombinedOutput()
 	if err != nil {
 		log.Errorf("nodes: failed to list vgpu profiles for gpu %s: %v", gpuId, err)
-		return &profiles
+		return collection
 	}
 
 	if !IsHexSuccessful(err) {
 		log.Errorf("nodes: output error when listing vgpu profiles for gpu %s via hex_sdk: %v", gpuId, err)
```
raven-pan: The err is already guaranteed nil in the previous `if err != nil {`.

So following code will never be executed.

```go
log.Errorf("nodes: output error when listing vgpu profiles for gpu %s via hex_sdk: %v", gpuId, err)
		return collection
```

internal/definition/v1/gpu/gpu.go
```diff
@@ -48,38 +55,43 @@ type GpuCard struct {
...
+}
+
 type VgpuProfile struct {
-	Id        string `json:"id"`
-	Name      string `json:"name"`
-	VramMiB   uint64 `json:"vramMiB"`
-	AliasName string `json:"aliasName"`
-	Count     int    `json:"count"`
-	Remaining int    `json:"remaining"`
+	Id         uint32  `json:"id"`
```
raven-pan, line 88: These field change from string to number.
Maybe we should update the openapi doc also?

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -420,20 +375,77 @@ func buildInstanceLinks(vmId string) gpu.InstanceLinks {
...
+	collection := gpu.GpuProfileCollection{
+		SriovVgpu:     []gpu.VgpuProfile{},
+		MigBackedVgpu: []gpu.VgpuProfile{},
 	}
 
-	profileInstanceCountMap := map[string]int{}
-	for _, instance := range attachedInstances {
-		profile := profileMapByAlias[*instance.ProfileAlias]
-		profileInstanceCountMap[profile.Id]++
+	for _, profile := range *hexProfileCollection.Sriov {
```
raven-pan: Maybe we should add a nil guard before `for _, profile := range *hexProfileCollection.Sriov {`?

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -420,20 +375,77 @@ func buildInstanceLinks(vmId string) gpu.InstanceLinks {
...
+			CountLimit: profile.VmCountLimit,
+		})
 	}
 
-	for _, profile := range profiles {
-		instanceCount := profileInstanceCountMap[profile.Id]
-		profile.Remaining = profile.Count - instanceCount
+	migProfileRemainingMap := createMigProfileRemainingMap(hexProfileCollection.MigBacked, attachedInstances)
+
+	for _, profile := range *hexProfileCollection.MigBacked {
```
raven-pan: Maybe we should add a nil guard before `or _, profile := range *hexProfileCollection.MigBacked {`?

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -104,7 +103,8 @@ func (h *helper) listLocalGpuCards() ([]gpu.GpuCard, error) {
 			return nil, err
 		}
 
-		vgpuProfiles, hexProfilesMap := listVgpuProfiles(device, hexGpu)
+		hexProfilesMap, hexProfileCollection := cubecos.GetNodeVgpuProfilesMap(hexGpu.PciAddress)
```
raven-pan: Discussion:
We can call `isVgpu` function before `cubecos.GetNodeVgpuProfilesMap(hexGpu.PciAddress)`.
If type is `unset` or `pgpu`, we can just skip the `cubecos.GetNodeVgpuProfilesMap(hexGpu.PciAddress)` call.

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -116,32 +116,29 @@ func (h *helper) listLocalGpuCards() ([]gpu.GpuCard, error) {
 			HexProfilesMap:           hexProfilesMap,
 		})
 
-		if vgpuProfiles != nil && attachedInstances != nil {
-			updateVgpuProfilesRemaining(*vgpuProfiles, *attachedInstances)
-		}
+		profileCollection := toProfileCollection(hexProfileCollection, attachedInstances)
```
raven-pan, line 124: Discussion:
We can call `isVgpu` function before `toProfileCollection`.
If type is `unset` or `pgpu`, we can just skip the `toProfileCollection` call and assign an empty array to `profileCollection.`

raven-pan (approved):

internal/definition/v1/gpu/gpu.go
steven-chiu-bigstack: Yes, and if I'm not mistaken, it should've been updated already (for list GPU cards API, see [docs](https://github.com/bigstack-oss/cube-cos-openapi/blob/221c16f26079be819bd357b9bf76f51b1f1724ff/docs.yaml#L17364)):

{168C33DD-DFFB-4AC9-8D62-E8B6E6E79DD1}

As for the openapi docs for updating GPU API, perhaps we can handle that during your API implementation?

internal/apis/v1/handlers/nodes/gpu.go
steven-chiu-bigstack: Yeah, you're right. Otherwise `profileId` will default to `0` thanks to GO's built-in behavior.

internal/apis/v1/handlers/nodes/gpu.go
steven-chiu-bigstack: No problem. 👌

internal/apis/v1/handlers/nodes/gpu.go
steven-chiu-bigstack: Sure, good catch!

internal/apis/v1/handlers/nodes/gpu.go
steven-chiu-bigstack: Sure, good catch!

internal/apis/v1/handlers/nodes/gpu.go
steven-chiu-bigstack: I prefer keeping the current pattern since it's pure data mapping (raw data to view model), which should happen regardless of the GPU type.

internal/cubecos/nodes.go
steven-chiu-bigstack: You're definitely right!
I copied this pattern from other files; not sure why there's a bunch of redundant check in the codebase. 🤔
I'll proceed to remove this `if !IsHexSuccessful(err) { ... }` block because it's unnecessary.

https://github.com/user-attachments/assets/c7ef175f-b435-436f-a88c-05bcd0fdef86

PR #602, opened 2026-07-03
refactor(gpu): fail fast on integrity errors, degrade enrichment failures via one accumulator

raven-pan:
### What type of PR is this?

Refactor / bug fix

<br/>

N/A

<br/>

### What this PR does?

Reworks error handling in the local GPU card listing. Previously errors were logged and swallowed, returning cards with silently missing data. Now there is an explicit two-tier policy: **integrity failures fail the request; enrichment failures degrade the affected card.**

**Integrity failures fail the request.** If the source of truth is broken, the listing errors out instead of fabricating partial data:

- hex_sdk GPU inventory and attached-instance lookups propagate errors to the handler response.
- `cubecos.GetNodePgpuAttachedInstance` returns `(nil, nil)` when no instance is attached and a wrapped error on hex_sdk/parse failures, instead of a zero-value instance indistinguishable from real data.
- Inconsistent hex data (allocation reported but no attached instance) — for the non-transient case — is an error.

**Enrichment failures degrade the card, funneled through one accumulator.** Runtime stats and links are extras on top of the hex inventory; one sick GPU or stopped VM must not hide every card on the node. Every "NVML/hex/Openstack enrichment could not be obtained" signal now routes through a single per-card `enrichment.degrade(reason...)` that couples the `degraded` flag with its log line — replacing the hand-copied `degraded=true` + log pairs and the bool threaded through `listAttachedInstances`/`listPgpu`/`listVgpu`. A `degraded` card means its capacity numbers are untrustworthy, so schedulers do not allocate off it.

- **Degrade (logs at error level — capacity untrustworthy):** device-handle fault; `GetMemoryInfo` / `GetUtilizationRates` faults; `GetActiveVgpus` fault (was a hard error that failed the whole listing); per-instance and profile-id NVML skips; `GetVgpuUtilization` non-SUCCESS (`NOT_SUPPORTED`, `GPU_IS_LOST`, …); hex vGPU-profile fetch failure (`GetNodeVgpuProfilesMap` now errors); pgpu attached-instance hex-lookup failure; `getOpenstackServerNames` failure (nil-map sentinel flags every card).
- **Soft, non-misleading gaps (log at warn, no degrade):** a single VM absent from a non-nil server map; transient pgpu alloc/instance read skew (allocation counted, instance not yet reported → no attached instance).
- NVML enrichment is tolerated as genuinely absent only on `ERROR_NOT_FOUND` (any vfio-bound GPU, including one reserved but not yet attached), logged at Debug — so real NVML faults are surfaced rather than misread as passthrough.
- Skip a stale vGPU instance handle instead of 500-ing the whole node listing; skip the utilization query entirely when there are no active vGPUs.
- Openstack server-name lookup failures report the instance without a name; a nil server no longer formats a `<nil>` error. A nil `Allocation` on a passthrough pGPU no longer panics.

**Console links are minted lazily via a dedicated endpoint.** Creating a Nova console is a *write* that mints a short-lived noVNC token; doing it for every attached instance on every GPU-list poll flooded Nova with sessions that are almost always discarded (a dashboard polling every few seconds on a node with several cards × several instances issued dozens of console POSTs per poll).

- The listing **no longer mints any console token**, and the `console` field is **removed from the attached-instance `links` object entirely** — not returned as a constant empty string, which would be a trap (an un-migrated client renders "no console" instead of calling the endpoint). Each instance's `links` now carries only the `grafana` dashboard URL.
- New endpoint **`GET /api/v1/datacenters/{dataCenter}/nodes/{nodeName}/gpuCards/instances/{instanceId}/console`** mints the Nova noVNC console on demand — only when the user opens a console — returning `{ "console": "<url>" }`, keeping the short-lived token fresh at click time.
- Attached-instance names are resolved with a **single `servers.List` per request** (prefetched into an id→name map) instead of a `GetServer` round trip per instance.

Also included:

- `fix(runtime)`: NVML lifecycle/availability moved into the `internal/nvmlruntime` leaf package. NVML init failure of any kind warns and continues without GPU support instead of `log.Fatalf`; the always-nil error return is dropped, "driver not loaded" is only claimed when NVML reports it, real init faults log at error level, and NVML is shut down gracefully on service stop.
- `fix(config)`: `flag.Parse` runs before the config file is loaded so a custom `-conf` path actually takes effect; flags still override file values, and the double-parse precedence is documented.
- `refactor(grafana)`: the Grafana instance-dashboard URL is built by the exported `grafana.InstanceDashboardLink`, deduplicating the copy in the GPU handler.
- `gpu_pgpu_attached_instance_get` reads stdout alone (`cmd.Output`) so a stderr warning cannot corrupt the JSON parse; `VALUE_TYPE_DOUBLE` utilization samples decode via `math.Float64frombits`.
- Unit tests for the GPU listing paths via package seams for hex_sdk, NVML, and Openstack (testify promoted to a direct dependency), plus tests for the grafana link helper and NVML init.

<br/>

### Test results (optional)

1). API docs: `task generateApiDocs` run as part of `task check`. The new on-demand console endpoint and the removal of the `console` field from the listing's attached-instance links are documented in the OpenAPI spec (bigstack-oss/cube-cos-openapi#106).

2). `task test` passes (GPU handler, grafana, and runtime unit tests included) and `task vet` is clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -365,19 +365,14 @@ func buildInstanceLinks(vmId string) gpu.InstanceLinks {
 	})
 	console, err := result.Extract()
 
-	consoleLink := ""
-
 	if console == nil || err != nil {
```
raven-pan: **[correctness / CONFIRMED]** Console-link creation failure for a non-ACTIVE VM now fails the entire GPU listing instead of leaving the console link empty.

A pGPU/vGPU attached to a VM in SHUTOFF/PAUSED/ERROR state: Nova rejects `remoteconsoles.Create` with 409 (consoles require ACTIVE instances), so GET node GPU cards returns 500 for the whole node — persistently, until the VM is started. Previously the card was returned with `Console:""`.

Also note: the endpoint creates a *new* noVNC console session per attached instance per poll — a dashboard polling every 5s mints thousands of console sessions per hour against Nova.

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -38,117 +49,109 @@ func (h *helper) listNodeGpuCards() ([]gpu.GpuCard, error) {
...
+func (h *helper) buildLocalGpuCard(hexGpu gpu.GpuFromHex) (gpu.GpuCard, error) {
+	var memoryUsedMiB, memoryTotalMiB int
+	var memoryUtilizationPercent, gpuUtilizationPercent uint32
+
+	device, ret := deviceGetHandleByUUID(hexGpu.Id)
+	isDeviceVisible := ret == nvml.SUCCESS
 
+	if !isDeviceVisible {
+		// Expected for a GPU passed through to a VM is invisible to NVML.
+		isPassedThroughToVm := hexGpu.Type == gpu.ResourceTypePgpu &&
```
raven-pan: **[correctness / CONFIRMED]** `isPassedThroughToVm` treats **any** NVML handle-lookup failure on an allocated pGPU as expected passthrough invisibility, silently reporting a faulted GPU as a healthy card.

An allocated pGPU whose handle lookup fails for a real fault — `ERROR_UNINITIALIZED` (NVML init swallowed an error at startup), `ERROR_GPU_IS_LOST` (card fell off the bus), `ERROR_UNKNOWN` — is reported as a healthy in-use passthrough card with zeroed VRAM/utilization, hiding a dead GPU from operators.

Fix: only treat `ret == nvml.ERROR_NOT_FOUND` as "invisible to NVML"; return an error for other codes.

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -228,111 +216,123 @@ func listPgpuAttachedInstances(opts listAttachedInstancesOpts) *[]gpu.AttachedIn
...
+	}
 
+	vgpuInstances, ret := device.GetActiveVgpus()
 	if ret != nvml.SUCCESS {
-		log.Errorf("nvml: failed to get active vgpus for device %s: %s", deviceUUID, nvml.ErrorString(ret))
-		return &attachedInstances
+		return nil, fmt.Errorf("nvml: failed to get active vgpus for device %s: %s", deviceUUID, nvml.ErrorString(ret))
+	}
+
+	vgpuInstanceUtilizationMap, err := buildVgpuInstanceUtilizationMap(device, deviceUUID)
```
raven-pan: **[correctness / PLAUSIBLE]** `buildVgpuInstanceUtilizationMap` is called unconditionally and now hard-fails on any non-SUCCESS from `GetVgpuUtilization`, but NVML returns `ERROR_NOT_FOUND` when there are no utilization samples (e.g. zero running vGPU VMs).

A GPU in SR-IOV/MIG vGPU mode with no running vGPU VMs, or right after driver start before the first sampling window: `GetActiveVgpus` returns an empty list but `GetVgpuUtilization(0)` returns `ERROR_NOT_FOUND`, and the whole GPU-cards endpoint 500s on an otherwise healthy node. The old code logged and used an empty map.

Fix: skip the call when `len(vgpuInstances) == 0` and tolerate `ERROR_NOT_FOUND`.

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -38,117 +49,109 @@ func (h *helper) listNodeGpuCards() ([]gpu.GpuCard, error) {
...
+	device, ret := deviceGetHandleByUUID(hexGpu.Id)
+	isDeviceVisible := ret == nvml.SUCCESS
 
+	if !isDeviceVisible {
+		// Expected for a GPU passed through to a VM is invisible to NVML.
+		isPassedThroughToVm := hexGpu.Type == gpu.ResourceTypePgpu &&
+			hexGpu.Allocation != nil &&
+			hexGpu.Allocation.Current > 0
+		if !isPassedThroughToVm {
+			return gpu.GpuCard{}, fmt.Errorf("nvml: failed to get device handle for gpu %s: %s", hexGpu.Id, nvml.ErrorString(ret))
```
raven-pan: **[correctness / PLAUSIBLE]** An idle pGPU that is invisible to NVML (vfio-bound for passthrough but with `Allocation` nil or `Current == 0`) now errors and kills the whole node listing.

A GPU reserved for passthrough is typically bound to vfio-pci at boot, so NVML cannot see it even with no VM attached; hex reports `Type=pgpu`, `Allocation.Current == 0`, `isPassedThroughToVm` is false, and every GET node GPU cards returns 500. The same window exists right after a passthrough VM is deleted (allocation decremented before the device is rebound to the NVIDIA driver).

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -38,117 +49,109 @@ func (h *helper) listNodeGpuCards() ([]gpu.GpuCard, error) {
...
-	for i := range deviceCount {
-		device, ret := nvml.DeviceGetHandleByIndex(i)
-		if ret != nvml.SUCCESS {
-			errorString := nvml.ErrorString(ret)
-			log.Errorf("nvml: failed to get device handle at index %d: %s", i, errorString)
-			continue
+	// Iterate over hex GPUs instead of NVML devices: a GPU passed through to
+	// a VM is invisible to NVML, but must still be reported.
+	for _, pciAddress := range slices.Sorted(maps.Keys(hexGpusMap)) {
+		gpuCard, err := h.buildLocalGpuCard(hexGpusMap[pciAddress])
```
raven-pan: **[correctness / CONFIRMED]** `listLocalGpuCards` aborts the entire listing on the first per-GPU error, so one sick GPU hides all GPUs on the node (and cluster-wide via `listRemoteGpuCards` proxying).

Node with 4 GPUs where one returns `ERROR_GPU_IS_LOST` from `GetMemoryInfo`: the old code returned 3 healthy cards plus hex data for the sick one; new code returns 500 for the node, and the control node's proxied listing fails too — operators lose all GPU visibility exactly when one card is failing.

Consider per-card degradation (e.g. an error field on the card) as handler-level policy instead of all-or-nothing.

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -228,111 +216,123 @@ func listPgpuAttachedInstances(opts listAttachedInstancesOpts) *[]gpu.AttachedIn
...
 		if ret != nvml.SUCCESS {
-			log.Errorf("nvml: failed to get fb usage for vgpu instance %s: %s", vmId, nvml.ErrorString(ret))
+			return nil, fmt.Errorf("nvml: failed to get fb usage for vgpu instance %s: %s", vmId, nvml.ErrorString(ret))
 		}
 
 		hexProfile := hexProfilesMap[profileId]
 		profileAlias := hexProfile.Alias
 
-		instanceName := ""
 		server, err := openstackHelper.GetServer(vmId)
```
raven-pan: **[correctness / PLAUSIBLE]** A failed or nil `openstackHelper.GetServer` for any single vGPU instance now fails the whole endpoint instead of leaving that instance's name blank.

A vGPU VM is deleted while NVML still reports its vGPU instance (teardown race), or Nova/Keystone is briefly unreachable: `GetServer` returns nil/404 and the whole node GPU listing 500s for the duration.

Minor: when `server == nil` with `err == nil`, the `%v` wrap produces `failed to get Openstack server vm-x: <nil>`.

internal/runtime/nvml.go
```diff
@@ -8,22 +8,39 @@ import (
...
+
 func initNvml() error {
 	ret := nvml.Init()
 	if ret != nvml.SUCCESS {
 		errorString := nvml.ErrorString(ret)
 
-		if ret == nvml.ERROR_DRIVER_NOT_LOADED {
-			log.Warnf("Failed to initialize NVML: %v", errorString)
-			log.Warnf("NVIDIA driver is not loaded. Continuing without NVML support, GPU features will be unavailable.")
+		log.Warnf("Failed to initialize NVML: %v", errorString)
```
raven-pan: **[correctness / CONFIRMED]** `initNvml` now treats every `nvml.Init` failure as "driver not loaded", logs a factually wrong message for real faults, and its error return is dead (both paths return nil, so the caller's `if err != nil` is unreachable).

After a partial driver upgrade (`ERROR_LIB_RM_VERSION_MISMATCH`) or with `ERROR_NO_PERMISSION`, the service starts "healthy" while logging 'NVIDIA driver is not loaded', and the fault surfaces later as confusing per-request 500s (or silently-zeroed passthrough cards via the `isPassedThroughToVm` interaction in gpu.go) instead of the old fail-fast startup crash.

Fix: keep graceful-continue only for `ERROR_DRIVER_NOT_LOADED`/`ERROR_LIBRARY_NOT_FOUND`, log the real error string for the rest.

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -38,117 +49,109 @@ func (h *helper) listNodeGpuCards() ([]gpu.GpuCard, error) {
...
-			continue
-		}
+	return gpuCards, nil
+}
+
+func (h *helper) buildLocalGpuCard(hexGpu gpu.GpuFromHex) (gpu.GpuCard, error) {
+	var memoryUsedMiB, memoryTotalMiB int
+	var memoryUtilizationPercent, gpuUtilizationPercent uint32
+
+	device, ret := deviceGetHandleByUUID(hexGpu.Id)
```
raven-pan, line 150: **[correctness / PLAUSIBLE]** The rewrite introduces an unverified cross-system contract that hex_sdk's gpu `id` equals the exact NVML `GPU-<uuid>` string; the old code correlated hex and NVML devices by (lowercased) PCI address via `extractPciAddress`.

If `hex_sdk gpu_device_list` emits an id in any non-NVML-UUID form (different scheme, case mismatch) for some GPU model, `deviceGetHandleByUUID` returns `ERROR_NOT_FOUND` for a visible idle GPU, and since `Allocation.Current == 0` the passthrough exemption doesn't apply — every GPU listing on that node 500s.

Worth verifying the hex id format guarantee on real hardware, or falling back to PCI-address correlation.

internal/config/config.go
```diff
@@ -158,6 +157,7 @@ func (o *Options) String() (string, error) {
 }
 
 func SyncOptions() error {
+	flag.Parse()
```
raven-pan, line 164: **[simplification / CONFIRMED]** `SyncOptions` calls `flag.Parse` twice — directly here and again via the `overrideOptsByFlags` wrapper — with nothing documenting the three-phase precedence dance (parse for `-conf`, file load clobbers `Opts`, re-parse restores CLI flags).

A future cleanup that removes the "redundant" second Parse silently makes the config file override explicit CLI flags; removing the first breaks `-conf`. No test covers either regression.

Suggest a comment explaining why parsing happens twice, or making precedence explicit with `flag.Visit` after the file load.

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -344,10 +344,10 @@ func buildVgpuInstanceUtilizationMap(device nvml.Device, deviceUUID string) map[
...
 	}
 
-	return utilizationMap
+	return utilizationMap, nil
 }
 
-func buildInstanceLinks(vmId string) gpu.InstanceLinks {
+func buildInstanceLinksViaOpenstack(vmId string) (gpu.InstanceLinks, error) {
 	grafanaLink := fmt.Sprintf(
 		"https://%s/grafana/d/PVW6vU7Wz/instance?refresh=5m&kiosk=tv&orgId=1&var-UUID=%s",
```
raven-pan: **[reuse / CONFIRMED]** This Grafana instance-dashboard URL template is byte-for-byte identical to the one in `internal/apis/v1/handlers/grafana/links.go:20`.

When the dashboard UID `PVW6vU7Wz` or query params change, GPU attached-instance links silently go stale because only links.go gets updated. Since this PR rewrote the enclosing function anyway, it's the natural moment to extract a shared link builder both packages call.

raven-pan (commented):
Automated code review (8 finder angles + verification). 10 findings below as inline comments — 8 correctness, 2 cleanup. Dominant theme: several error paths hardened by the fail-fast conversion are routine conditions rather than faults, so they now 500 the whole GPU listing. Highest-leverage fix: keep hard errors for integrity problems (hex_sdk/JSON), but let enrichment failures (console link, server name, utilization samples) degrade per-card.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Fixed in f8c35287: console creation moved behind a `createConsole` seam and `buildInstanceLinksViaOpenstack` now degrades — on failure (e.g. Nova 409 for a stopped VM) it logs a warning and returns the Grafana link with an empty console link.

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Fixed in f8c35287: the allocation-based `isPassedThroughToVm` guess is gone. Tolerance is keyed on the NVML return code — `ERROR_NOT_FOUND` is treated as an expected vfio-bound GPU, while any other fault is logged as a warning and the card is reported without runtime stats.

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Fixed in f8c35287: `GetVgpuUtilization` returning `ERROR_NOT_FOUND` (no samples yet) now yields an empty utilization map instead of an error.

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Fixed in f8c35287: any `ERROR_NOT_FOUND` from the handle lookup is tolerated regardless of allocation state, so an idle vfio-reserved pGPU is reported from hex data without NVML stats.

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Addressed in f8c35287: NVML/OpenStack enrichment failures (handle, memory, utilization, console link, server name) now degrade per card/instance with warnings instead of aborting. Hex inventory failures still fail the request, as they indicate the source of truth itself is broken.

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Fixed in f8c35287: the server lookup is seamed and degrades — on error or nil server the instance is reported with an empty name and a warning is logged (the nil case gets its own message, so no more `<nil>` formatting).

internal/runtime/nvml.go
raven-pan: Fixed in 972e542d: `initNvml` no longer returns an (always-nil) error, and the "driver not loaded" message is only logged for `ERROR_DRIVER_NOT_LOADED` / `ERROR_LIBRARY_NOT_FOUND`; other failures log the actual NVML error string.

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Acknowledged in f8c35287 with a comment documenting the contract. With the ERROR_NOT_FOUND handling, a UUID mismatch now degrades to a card without runtime stats instead of failing the listing, so the blast radius of a broken contract is a warning log rather than a 500.

internal/config/config.go
raven-pan: Fixed in eb8af7d1: `SyncOptions` and `overrideOptsByFlags` now document the flag > file > default precedence and why `flag.Parse` runs twice.

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Fixed in c938adc9: the URL is now built by the exported `grafana.InstanceDashboardLink`, used by both handlers.

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -228,91 +221,100 @@ func listPgpuAttachedInstances(opts listAttachedInstancesOpts) *[]gpu.AttachedIn
...
 
-	openstackHelper := openstack.GetGlobalHelper()
+	vgpuInstanceUtilizationMap, err := buildVgpuInstanceUtilizationMap(device, deviceUUID)
+	if err != nil {
+		return nil, err
+	}
 
 	for i, vgpuInstance := range vgpuInstances {
 		vmId, _, ret := vgpuInstance.GetVmID()
 		if ret != nvml.SUCCESS {
```
raven-pan, line 446: **[correctness / CONFIRMED]** Per-vGPU-instance NVML calls (`GetVmID`/`GetType`/`GetGpuInstanceProfileId`/`GetFramebufferSize`/`GetFbUsage`) hard-fail the whole node listing, contradicting this PR's own degrade-enrichment policy one layer down. A VM torn down between `GetActiveVgpus` and `GetVmID` returns `ERROR_NOT_FOUND` on the now-stale handle, the error propagates through `buildLocalGpuCard` → `listLocalGpuCards`, and `GET .../gpus` 500s for every GPU on the node during each teardown window (old code skipped just that instance).

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -323,16 +325,21 @@ func listVgpuAttachedInstances(opts listAttachedInstancesOpts) *[]gpu.AttachedIn
...
+	// NVML returns ERROR_NOT_FOUND when no utilization samples exist (e.g.
+	// no vGPU has been active since the last query); that is not a failure.
+	if ret == nvml.ERROR_NOT_FOUND {
+		return utilizationMap, nil
+	}
+
 	if ret != nvml.SUCCESS {
-		log.Errorf("nvml: failed to get vgpu utilization for device %s: %s", deviceUUID, nvml.ErrorString(ret))
-		return utilizationMap
+		return nil, fmt.Errorf("nvml: failed to get vgpu utilization for device %s: %s", deviceUUID, nvml.ErrorString(ret))
```
raven-pan: **[correctness / CONFIRMED]** `buildVgpuInstanceUtilizationMap` only tolerates `ERROR_NOT_FOUND`; NVML also documents `ERROR_NOT_SUPPORTED` and `ERROR_GPU_IS_LOST` for `GetVgpuUtilization`. `NOT_SUPPORTED` makes the endpoint fail **permanently** on some hardware, and `GPU_IS_LOST` during a reset fails it transiently — where the pre-refactor code degraded to an empty map. Suggest tolerating any non-SUCCESS here (return an empty map + warn).

internal/runtime/nvml.go
```diff
@@ -8,22 +8,40 @@ import (
...
+
+// NVML is optional (nodes without NVIDIA GPUs run the same service), so any
+// init failure degrades to running without NVML support instead of aborting.
+func initNvml() {
+	switch ret := nvmlInit(); ret {
+	case nvml.SUCCESS:
+		isNvmlInitialized = true
+	case nvml.ERROR_DRIVER_NOT_LOADED, nvml.ERROR_LIBRARY_NOT_FOUND:
+		log.Warnf("NVIDIA driver is not loaded. Continuing without NVML support.")
+	default:
```
raven-pan: **[correctness / CONFIRMED]** A swallowed NVML init failure serves wrong data with HTTP 200 indefinitely. After a transient boot failure, every `deviceGetHandleByUUID` returns `ERROR_UNINITIALIZED` (the default warn+degrade branch), so a MIG-backed card with 2 attached VMs reports `attachedInstances=[]` and `Remaining=full Count` — schedulers over-allocate off wrong capacity until a manual restart. Nothing in the handler layer consults `isNvmlInitialized`. Old code `log.Fatalf`'d on non-driver-missing failures, so this state was unreachable. Consider retrying init, or exposing availability so the handler can signal degraded mode.

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -228,91 +221,100 @@ func listPgpuAttachedInstances(opts listAttachedInstancesOpts) *[]gpu.AttachedIn
...
-			Name:               attachedInstance.Name,
-			ProfileAlias:       nil,
-			UtilizationPercent: deviceGpuUtilizationRate,
-			MemoryUsage: gpu.InstanceMemoryUsage{
-				AllocatedMiB: deviceMemoryUsedMiB,
-				TotalMiB:     deviceMemoryTotalMiB,
-			},
-			Links: buildInstanceLinks(attachedInstance.Id),
-		})
+		return nil, fmt.Errorf("gpu: hex gpu %s allocation.current is not 0, but its attached instance is null on node %s", hexGpu.PciAddress, nodeName)
```
raven-pan: **[correctness / PLAUSIBLE]** A transient hex read-skew (`Allocation.Current=1` while `gpu_pgpu_attached_instance_get` returns no instance) is now a hard error that 500s the whole node listing for the duration of the race window. During normal pGPU attach/detach the two hex queries are read at different instants; `GetNodePgpuAttachedInstance` returns `(nil, nil)` while `Current` is still 1, and every card on the node disappears from the API until hex converges. Old code logged and returned the card with an empty instance list.

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -38,117 +53,110 @@ func (h *helper) listNodeGpuCards() ([]gpu.GpuCard, error) {
...
-		hexGpu := hexGpusMap[pciAddress]
-
-		if len(hexGpu.Id) == 0 {
-			err := fmt.Errorf("Cannot find hex gpu with pci address %s", pciAddress)
-			log.Errorf(err.Error())
-			return nil, err
+			log.Warnf("nvml: failed to get utilization rates for device %s: %s", hexGpu.Id, nvml.ErrorString(ret))
 		}
+	case ret == nvml.ERROR_NOT_FOUND:
+		// Expected: a GPU bound to vfio for passthrough (attached to a VM or
```
raven-pan: **[correctness / PLAUSIBLE]** This `ERROR_NOT_FOUND` branch is completely silent, so any mismatch between hex's GPU `id` and the exact NVML `GPU-<uuid>` string silently strips all runtime stats and vGPU attached instances from every card at HTTP 200 with zero log signal. The old code correlated hex↔NVML by PCI address, so this exact-UUID contract is unverified. A `Debugf`/metric here, or a one-time startup cross-check of hex ids against NVML UUIDs, would make the regression visible.

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -38,117 +53,110 @@ func (h *helper) listNodeGpuCards() ([]gpu.GpuCard, error) {
...
 
-	for i := range deviceCount {
-		device, ret := nvml.DeviceGetHandleByIndex(i)
-		if ret != nvml.SUCCESS {
-			errorString := nvml.ErrorString(ret)
-			log.Errorf("nvml: failed to get device handle at index %d: %s", i, errorString)
-			continue
+	// Iterate over hex GPUs instead of NVML devices: a GPU passed through to
+	// a VM is invisible to NVML, but must still be reported.
+	for _, pciAddress := range slices.Sorted(maps.Keys(hexGpusMap)) {
```
raven-pan, line 102: **[correctness / CONFIRMED]** A GPU visible to NVML but absent from hex inventory now silently disappears from the response. The hex-driven loop never reconciles against NVML enumeration; the old code hard-errored (`Cannot find hex gpu with pci address ...`) to surface exactly this mismatch. When `hex_sdk gpu_device_list` lags hardware (hot-add, stale cache, registration bug), operators lose the signal the old error path gave them.

internal/cubecos/nodes.go
```diff
@@ -463,32 +463,26 @@ func getNodeVgpuProfileCollection(gpuId string) gpu.VgpuProfileCollectionFromHex
...
-		return &attachedInstance
-	}
-
-	if !IsHexSuccessful(err) {
-		log.Errorf("nodes: output error when getting attached instance for gpu %s via hex_sdk: %v", pciAddress, err)
-		return &attachedInstance
+		return nil, fmt.Errorf("nodes: failed to get attached instance for gpu %s via hex_sdk: %w", pciAddress, err)
 	}
 
 	err = json.Unmarshal(out, &attachedInstance)
```
raven-pan, line 485: **[correctness / CONFIRMED]** `GetNodePgpuAttachedInstance` unmarshals `CombinedOutput` (stdout+stderr merged) and now returns a hard error on parse failure, which 500s the whole GPU listing. If `hex_sdk gpu_pgpu_attached_instance_get` prints a warning line to stderr ahead of the JSON, the unmarshal fails for every GPU on the node. The old code logged and returned an empty struct. Capture stdout alone via `cmd.Output()`, or degrade the parse failure.

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -344,16 +351,32 @@ func buildVgpuInstanceUtilizationMap(device nvml.Device, deviceUUID string) map[
 		}
```
raven-pan, line 527: **[correctness / PLAUSIBLE]** `VALUE_TYPE_DOUBLE` samples are decoded with `binary.LittleEndian.Uint32` over the raw bytes of an IEEE-754 double, yielding 0 or garbage utilization (pre-existing line, now load-bearing in the rewritten per-instance path). A driver returning `33.0` as a double gives `Uint32` of the low mantissa bytes = 0. Reinterpret via `math.Float64frombits` and round instead.

raven-pan (commented):
Second-pass automated code review (8 finder angles + verification, high effort). 10 findings — the theme is that the degrade-enrichment policy from the last round stops one layer too early, and two failure modes are now *silent* instead of loud. Ranked most-severe first; each inline comment is tagged with category and verdict (CONFIRMED / PLAUSIBLE).

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -363,21 +386,18 @@ func buildInstanceLinks(vmId string) gpu.InstanceLinks {
 		Protocol: remoteconsoles.ConsoleProtocolVNC,
```
raven-pan, line 567: **[efficiency / CONFIRMED]** Every GPU list request performs a Nova remote-console **CREATE** (a write minting a console token, 10s timeout) plus a `GetServer` round trip per attached instance, all sequentially, discarding the token unless the user opens a console. A dashboard polling every 10s on a node with 4 cards × 4 instances issues 16 console POSTs + 16 GetServer calls per poll. Build the console link lazily via a dedicated endpoint (or cache per `vmId` for the token TTL), and resolve names with one `servers.List` per request.

internal/apis/v1/handlers/nodes/gpu.go
```diff
@@ -228,91 +221,100 @@ func listPgpuAttachedInstances(opts listAttachedInstancesOpts) *[]gpu.AttachedIn
...
+	vgpuInstances, ret := device.GetActiveVgpus()
 	if ret != nvml.SUCCESS {
-		log.Errorf("nvml: failed to get active vgpus for device %s: %s", deviceUUID, nvml.ErrorString(ret))
-		return &attachedInstances
+		return nil, fmt.Errorf("nvml: failed to get active vgpus for device %s: %s", deviceUUID, nvml.ErrorString(ret))
 	}
 
-	openstackHelper := openstack.GetGlobalHelper()
+	vgpuInstanceUtilizationMap, err := buildVgpuInstanceUtilizationMap(device, deviceUUID)
+	if err != nil {
```
raven-pan: **[efficiency / CONFIRMED]** `buildVgpuInstanceUtilizationMap` runs an NVML driver query (whose failure can fail the card — see the `GetVgpuUtilization` finding above) even when `GetActiveVgpus` just returned zero instances, and the result is never read. Guard with `if len(vgpuInstances) == 0 { return &attachedInstances, nil }` before building the map — also shrinks the blast radius of the `NOT_SUPPORTED` case.

steven-chiu-bigstack (approved):
LGTM!

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Fixed in d6a16908: the per-instance NVML calls (GetVmID/GetType/GetGpuInstanceProfileId/GetFramebufferSize/GetFbUsage) now log a warning and `continue` past just that instance, so a stale handle from a VM torn down between GetActiveVgpus and these calls no longer 500s the whole node listing.

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Fixed in d6a16908: buildVgpuInstanceUtilizationMap now degrades on any non-SUCCESS return (NOT_SUPPORTED, GPU_IS_LOST, NOT_FOUND, ...), logging a warning and returning an empty map instead of erroring.

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Fixed in d6a16908: listVgpuAttachedInstances now returns early when GetActiveVgpus reports zero instances, so the utilization query is skipped entirely — which also shrinks the blast radius of the NOT_SUPPORTED case above.

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Fixed in d6a16908: a nil attached instance while allocation.current > 0 (transient hex read skew) now logs a warning and returns no attached instance, instead of erroring and 500-ing the node.

internal/cubecos/nodes.go
raven-pan: Fixed in d6a16908: GetNodePgpuAttachedInstance now uses cmd.Output() (stdout only) instead of CombinedOutput(), so a warning line printed to stderr ahead of the JSON can no longer corrupt the unmarshal.

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Fixed in d6a16908: VALUE_TYPE_DOUBLE samples are now reinterpreted via math.Float64frombits(binary.LittleEndian.Uint64(...)) and rounded, rather than reading the low bytes as a little-endian Uint32.

internal/runtime/nvml.go
raven-pan: Addressed in d6a16908: NVML lifecycle + availability moved into the internal/nvmlruntime leaf package, and a real init fault now logs at Error (not the "driver not loaded" message). IsAvailable() is exposed to the handler, and each GpuCard now carries a `degraded` flag set when NVML enrichment it should have had is missing (e.g. deviceGetHandleByUUID returns ERROR_UNINITIALIZED after a swallowed init failure). So the wrong-capacity state now surfaces as `degraded: true` at HTTP 200 rather than silently. The flag is per-card so it survives the control-node proxy; happy to switch to a node/envelope-level signal if you prefer.

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Fixed in d6a16908: after building cards from hex, warnGpusMissingFromHex enumerates NVML devices (DeviceGetCount + DeviceGetHandleByIndex + GetUUID) and warns for any GPU visible to NVML but absent from the hex inventory. It runs only while NVML is available, since a vfio-passthrough GPU is legitimately invisible to NVML and would otherwise warn falsely.

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Addressed in d6a16908: the expected passthrough ERROR_NOT_FOUND branch now logs a Debugf, and the new warnGpusMissingFromHex reconciliation (see the finding below) surfaces a hex-id vs NVML-UUID mismatch as a warning at listing time, so the exact-UUID contract is no longer enforced entirely silently.

internal/apis/v1/handlers/nodes/gpu.go
raven-pan: Fixed in d6a16908: console links are no longer minted during the listing — the inline `console` field is now always empty. A new `GET .../gpuCards/instances/{instanceId}/console` endpoint mints the Nova console on demand, and vGPU instance names are resolved with a single ListServers per request instead of a GetServer per instance. OpenAPI documented in bigstack-oss/cube-cos-openapi#106.

PR #617, opened 2026-07-24
fix(time): emit local timezone for every mislabeled timestamp, drop ISO8601Z (#1166)

### LINKED ISSUE #1166: [Bug]: Fixpack "last updated" timestamp displayed 8 hours in the future
## Describe the bug

The Maintenance → Update → Fixpack page (`/maintenance/update/fixpack`) shows the
"last time" for an installed fixpack as a timestamp roughly 8 hours in the future
relative to when it was actually installed. E.g. fixpack `v3.1.0-002` was actually
installed at `2026-07-23 19:02:46` (CST, confirmed via `hex_cli -c fixpack
view_history` on multiple cluster nodes), but the web page displays
`2026-07-24 03:02:46`.
```bash
c1> fixpack view_history
Fix Packs History:
Date                  Id         Title           Action       Rollback Description
23 Jul 2026 18:51:51  001        appctl 4eaec97  Installed    No       fix appfw offline install
23 Jul 2026 19:02:46  v3.1.0-002 v3.1.0-002      Installed    Yes      Security fixes: kernel 6.12.95-1.el9.x86_64
```

Image

## Affected Component / Service

`cube-cos-api` — specifically `internal/cubecos/datacenter.go`,
`GetFixpackUpdatedAt()`.

## Version

- Product/Fixpack version: CubeCOS fixpack v3.1.0-002
- Component version (if different from product version): cube-cos-api as deployed
  on a 3-control + 2-compute HA cluster (VIP 10.32.36.10), observed 2026-07-24

## To Reproduce

1. Install any fixpack (e.g. `v3.1.0-002`) on a cluster whose system timezone is not
   UTC (this cluster runs CST, UTC+8).
2. Confirm the real install time via CLI: `hex_cli -c fixpack view_history` — shows
   the correct local time (e.g. `23 Jul 2026 19:02:46`).
3. Open the web UI's Maintenance → Update → Fixpack page (backed by the
   `cube-cos-api` REST API), and look at the "last time" field for that fixpack.
4. Observe it displays a time 8 hours ahead of the value from step 2.

## Expected behavior

The "last time" shown on the web page should match the actual local installation
time reported by `hex_cli -c fixpack view_history` (e.g. `2026-07-23 19:02:46`),
not a value shifted 8 hours into the future.

## Root Cause

`GetFixpackUpdatedAt()` in `internal/cubecos/datacenter.go` parses the raw
`hex_config fixpack_get_history` timestamp string with plain `ostime.Parse`:

```go
t, err := ostime.Parse(time.FormatFixpack, updatedAt)
...
return time.LocalRFC3339(t), nil
```

`time.FormatFixpack` (`"02 Jan 2006 15:04:05"`) has no timezone indicator, and Go's
`time.Parse` documents that in this case the result is interpreted as **UTC**. So a
raw local wall-clock string like `"23 Jul 2026 19:02:46"` (actually CST/+8) gets
parsed as if it were `2026-07-23T19:02:46Z` — 8 hours later than the true instant.

`time.LocalRFC3339(t)` then does `t.In(LocalFixedZone)`, converting that
already-wrong absolute instant into the cluster's local (+8) wall-clock
representation: `19:02:46 UTC + 8h = 2026-07-24 03:02:46`. The two steps compound
into exactly the observed +8h drift.

Confirmed by comparison with the sibling helper `convertRawTime()` in
`internal/cubecos/firmware.go`, which parses the same kind of raw local-time string
correctly:

```go
func convertRawTime(layout, rawTime string) string {
	t, err := ostime.ParseInLocation(layout, rawTime, time.LocalFixedZone)
	...
	return time.RFC3339Z(t)
}
```

`ParseInLocation` tells Go to interpret the zone-less input as already being in
`LocalFixedZone`, avoiding the UTC-default assumption — this is why the fixpack
**list** view's per-row timestamps display correctly, while this specific
"last operation" summary field does not.

## Suggested Fix

In `GetFixpackUpdatedAt()`, replace:
```go
t, err := ostime.Parse(time.FormatFixpack, updatedAt)
```
with:
```go
t, err := ostime.ParseInLocation(time.FormatFixpack, updatedAt, time.LocalFixedZone)
```
matching the pattern already used in `convertRawTime()`.

## Additional Context

- Cluster topology: 3 control + 2 compute HA, VIP 10.32.36.10
- Hardware: Dell PowerEdge R630-class, system timezone CST (UTC+8)
- Supporting logs:
  ```
  $ hex_cli -c fixpack view_history
  Date                  Id         Title           Action       Rollback Description
  23 Jul 2026 19:02:46  v3.1.0-002 v3.1.0-002      Installed    Yes      Security fixes: kernel 6.12.95-1.el9.x86_64
  ```
  (confirmed identical on both `c1` and `c3` controller nodes)

- [x] **Handbook knowledge update** — land the durable, team-readable knowledge from this

🤖 Generated with [Claude Code](https://claude.com/claude-code)

raven-pan:
### What type of PR is this?

`bug` — timezone correctness fix

<br/>

Fixes bigstack-oss/cubecos#1166

Handbook write-up: bigstack-oss/bigstack-handbook#154 — `kb/cubecos/architecture/api-response-timestamp-timezone.md`, the timestamp timezone contract plus an audit of every timestamp path.

<br/>

### What this PR does?

Every timestamp cube-cos-api returns must carry the node's local UTC offset — the node's timezone **is** the datacenter's timezone. Five response fields did not. All are fixed here, and the helper that caused most of them is removed so the mistake cannot recur.

**The two root causes**

1. **Double shift.** A zone-less CLI layout (`FormatFixpack`, `FormatBmc`) parsed with `time.Parse` is read as UTC; `LocalRFC3339` then applies `.In(LocalFixedZone)` on top, moving the clock by the node offset *and* labeling it local. Wrong digits, wrong instant.
2. **Fake zone label.** `FormatISO8601Z = "2006-01-02T15:04:05+00:00"`. In a Go layout `+00:00` is a literal, not a zone token, so `ISO8601Z` emitted local wall-clock digits asserted to be UTC — off by the whole node offset for any client that parses it.

**Fields fixed**

| Field | Was | Now |
|---|---|---|
| `GET /datacenters/:DataCenter` → `fixpack.updatedAt` | double shift | `ParseInLocation` + `RFC3339Z` |
| Support file `createdAt` (2 call sites) | fake `+00:00` | `LocalRFC3339` |
| FRU `manufacturingDate` | both causes at once | `convertBmcTime` → `ParseInLocation` + `RFC3339Z` |
| Trigger `status.updatedAt` (both branches) | fake `+00:00` | `LocalRFC3339` |
| Tunings `UpdatedAt` (via `time.Boot()`) | fake `+00:00` | `LocalRFC3339` |

**Cleanup**

With `Boot()` converted, nothing reached the broken helpers, so they are gone: `FormatISO8601Z`, `ISO8601Z`, `TimeISO8601Z`, a second unreferenced copy of the same literal in the triggers package, and the unused `FormatISO8601`. No format constant left in the package silently mislabels a zone.

**Checked, no change needed**

- `GetActiveFirmwaretUpdatedAt` (`time.Unix` + `LocalRFC3339`), the firmware list and fixpack list (`convertRawTime`), and the license dates (inlined `LocalRFC3339`) already emit correct local time.

**Out of scope — found by widening the audit afterwards**

The audit behind this PR followed `ISO8601Z` and the zone-less CLI layouts. Grepping *every* emit helper in the `time` package for callers found two more things, both deliberately left out of this PR:

- **bigstack-oss/cubecos#1182** — the Cinder storage paths. `GET /integrations/storages/:name` never converts the timezone at all (returns raw OpenStack UTC), while the list endpoint returns the same storage in local offset; and the list's empty-value fallback fills in `NowRFC3339()`, which both emits `Z` and fabricates request time so the value changes on every `GET`. Left out because the fix changes behaviour, not just a format — a drive-by zone swap here would hide the fabrication while making the path look audited.
- **bigstack-oss/cubecos#1183** — `LocalFixedZone` is a `FixedZone` snapshot taken at startup, so it goes stale across a DST transition until the process restarts, and `NowLocal()` reaches the offset through Go's DST-aware runtime zone instead. No current deployment is affected (no CST/+08:00 region observes DST); filed as the blocker to clear before any deployment in a DST region. Left out because it is a design decision touching every timestamp path.

So this PR closes #1166's two bug classes and removes the helper behind them. It does not make every timestamp in the codebase contract-clean — #1182 is the remaining live gap.

<br/>

### Test results (optional)

No API doc change — response field types and paths are unchanged. Only the timezone offset inside existing `updatedAt` / `createdAt` / `manufacturingDate` strings differs.

<br/>

Tests added for every fixed path — previously these packages had none. Each is tabled over **+08:00, -05:00 and UTC**, driven through stubbed seams rather than `hex_config`, IPMI, or a live request.

Both bug classes are **invisible at offset 0** — every UTC case passes even with the fix reverted — so single-zone tests would not have caught either. Each test was confirmed to fail before the fix was restored:

| Case | Expected | Actual with bug |
|---|---|---|
| fixpack @ +08:00 | `2025-07-24T15:02:37+08:00` | `2025-07-24T23:02:37+08:00` |
| fixpack @ −05:00 | `2025-07-24T15:02:37-05:00` | `2025-07-24T10:02:37-05:00` |
| support file / trigger / boot offset | `28800` | `0` |

- `go vet ./...` clean
- `go test -count=1 ./...` passes across all 9 packages
- Squashed to a single commit, rebased onto the current `develop` tip, and vet/test re-run on that tree

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/apis/v1/handlers/nodes/ipmi.go
+++ b/internal/apis/v1/handlers/nodes/ipmi.go
@@ -26,12 +26,23 @@ func (h *helper) verifyNodeIpmi() (*ipmi.FRU, error) {
 		return nil, fmt.Errorf("unable to connect with IPMI, please check your IPMI settings")
 	}
 
-	dateTime, err := time.Parse(bstime.FormatBmc, fru.ManufacturingDate)
+	manufacturingDate, err := convertBmcTime(fru.ManufacturingDate)
 	if err == nil {
-		fru.ManufacturingDate = bstime.ISO8601Z(dateTime)
+		fru.ManufacturingDate = manufacturingDate
 	}
 
 	return fru, nil
+}
+
+// FormatBmc carries no zone token, so the BMC wall clock must be parsed as the
+// node's local time — otherwise it is read as UTC and mislabeled.
+func convertBmcTime(rawTime string) (string, error) {
+	t, err := time.ParseInLocation(bstime.FormatBmc, rawTime, bstime.LocalFixedZone)
+	if err != nil {
+		return "", err
+	}
+
+	return bstime.RFC3339Z(t), nil
 }
 
 func (h *helper) checkBoardSerialConsistency(fru *ipmi.FRU) error {
--- a/internal/apis/v1/handlers/nodes/ipmi_test.go
+++ b/internal/apis/v1/handlers/nodes/ipmi_test.go
@@ -1,0 +1,59 @@
+package nodes
+
+import (
+	"testing"
+	ostime "time"
+
+	bstime "github.com/bigstack-oss/cube-cos-api/internal/definition/v1/time"
+	"github.com/stretchr/testify/require"
+)
+
+// Pins the process-wide local zone the way runtime.initSystemTime would, and
+// restores it so the offset does not leak between tests.
+func useLocalZone(t *testing.T, offsetSeconds int) {
+	t.Helper()
+
+	origSeconds := bstime.LocalZoneSeconds
+	origZone := bstime.LocalFixedZone
+
+	bstime.LocalZoneSeconds = offsetSeconds
+	bstime.LocalFixedZone = ostime.FixedZone("", offsetSeconds)
+
+	t.Cleanup(func() {
+		bstime.LocalZoneSeconds = origSeconds
+		bstime.LocalFixedZone = origZone
+	})
+}
+
+// The BMC reports a bare wall clock with no zone. It is the node's local time,
+// so the emitted value must carry the node's offset and keep the same digits.
+func TestConvertBmcTimeKeepsTheWallClockAndAddsTheLocalOffset(t *testing.T) {
+	tests := []struct {
+		name          string
+		offsetSeconds int
+		want          string
+	}{
+		{name: "east of UTC", offsetSeconds: 8 * 60 * 60, want: "2025-07-24T15:02:37+08:00"},
+		{name: "west of UTC", offsetSeconds: -5 * 60 * 60, want: "2025-07-24T15:02:37-05:00"},
+		{name: "at UTC", offsetSeconds: 0, want: "2025-07-24T15:02:37Z"},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			useLocalZone(t, test.offsetSeconds)
+
+			got, err := convertBmcTime("Thu Jul 24 15:02:37 2025")
+
+			require.NoError(t, err)
+			require.Equal(t, test.want, got)
+		})
+	}
+}
+
+func TestConvertBmcTimeRejectsAnUnparseableWallClock(t *testing.T) {
+	useLocalZone(t, 8*60*60)
+
+	_, err := convertBmcTime("Unspecified")
+
+	require.Error(t, err)
+}
--- a/internal/apis/v1/handlers/supportfiles/createdat_test.go
+++ b/internal/apis/v1/handlers/supportfiles/createdat_test.go
@@ -1,0 +1,104 @@
+package supportfiles
+
+import (
+	nethttp "net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	ostime "time"
+
+	"github.com/bigstack-oss/cube-cos-api/internal/definition/v1/status"
+	"github.com/bigstack-oss/cube-cos-api/internal/definition/v1/support"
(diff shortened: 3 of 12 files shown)


steven-chiu-bigstack (approved):
LGTM!

PR #620, opened 2026-07-27
fix(gpu): drop NVML and fix vGPU profile reporting

SekiXu:
### What type of PR is this?

/kind bug

Fixes #618 (`[Bug]: cube-cos-api permanently holds NVML, blocking SR-IOV vGPU unbindLock`)

Commits 2 and 3 have no ticket — both were found during the hardware
verification for #618 and are pre-existing, unrelated to NVML.

### What this PR does?

Three commits. The first is the #618 fix; the other two are pre-existing bugs
on the vGPU profile path that the same verification run uncovered. Please
review commit by commit.

---

**Commit 1 — drop NVML, source GPU stats from nvidia-smi.**

`nvml.Init()` runs at startup and `nvml.Shutdown()` is never called, so the
process holds an NVML device handle for its entire lifetime. That handle
permanently blocks the NVIDIA driver's unbindLock, which means **any** SR-IOV
or MIG vGPU management operation (`sriov-manage -e` / `-d`) fails on every
node running this service.

A lazy per-request Init/Shutdown fix was prototyped and hardware-verified
(branch `fix/nvml-lazy-init-unbind-lock`, not part of this change), but the
team chose to remove the in-process NVML dependency entirely instead:
nvidia-smi is a subprocess that starts, queries, and exits, so it
structurally cannot hold the handle open between requests — no Init/Shutdown
discipline to get wrong, no lock window to reopen under a slow request.

`internal/cubecos/gpu_nvidiasmi.go` wraps three invocations behind the same
CLI-wrapping convention CLAUDE.md documents:

- `nvidia-smi -q` — once per listing request, all devices — per-device UUID /
  memory / utilization.
- `nvidia-smi vgpu -q` — once per request when hex reports any vGPU GPU — for
  active vGPU instances, grouped by PCI bus id.
- `nvidia-smi vgpu -s -v -i <pci>` — lazily, MIG-backed cards only — to
  resolve a vGPU Type ID to its GPU Instance Profile ID. hex's MIG-backed
  profile ids are the GI Profile ID, not the vGPU Type ID that `vgpu -q`
  reports per instance, so this second lookup is required for MIG (SR-IOV's
  hex profile id is the vGPU Type ID directly).

`internal/nvmlruntime`, `internal/runtime/nvml.go`, and the `go-nvml` module
dependency are deleted outright rather than deprecated.

Two deliberate behaviour changes fall out of the atomic per-request
subprocess snapshot: per-field degrade collapses into one atomic per-device
degrade (nvidia-smi returns one snapshot per device, not separate fallible
calls), and the old "vGPU instance handle went stale mid-enumeration"
race/skip path is gone entirely.

Also fixed while writing the new degrade tests: the switch originally checked
"is this a pgpu" before "is nvidia-smi even available", which would have
silently misread a node-wide nvidia-smi outage as "every pgpu happens to be
expected vfio-passthrough" instead of degrading.

---

**Commit 2 — fractional `vramMiB` from hex_sdk broke the decode.**

`hex_sdk gpu_vgpu_profile_list` derives a MIG-backed profile's `vramMiB` from
the GiB column of `nvidia-smi mig -lgip` (`GiB * 1024`, `printf "%.2f"`), so
it reports a fractional MiB — e.g. **23674.88** for the 23.12 GiB "1g.24gb"
profile on an RTX PRO 6000 Blackwell. `VgpuProfileFromHex.VramMiB` was
`uint64`, which cannot decode that:

```
ERRO nodes: failed to parse output when listing vgpu profiles for gpu
00000001:C8:00.0 via hex_sdk: json: cannot unmarshal number 23674.88 into Go
struct field VgpuProfileFromHex.migBacked.vramMiB of type uint64
```

`encoding/json` does not stop at that error — it keeps decoding and leaves
only the offending field at its zero value. So the failure is silent in the
data and loud only in the log: every MIG-backed profile came back with
`vramMiB: 0`, a value that reads as legitimate. Confirmed live on cn13, where
all 11 MIG-backed profiles reported 0 while the SR-IOV ones decoded normally.

Two consequences, both worse than a visible parse failure:

- The card is flagged `degraded` (the returned error is non-nil), so its
  capacity is treated as untrustworthy for as long as the hardware has a MIG
  catalog with non-integer VRAM sizes — permanently, in practice.
- `validateGpuCardUpdate` sums `int(profile.VramMiB) * count` into
  `totalVramReqMiB` and rejects a migBackedVgpu request exceeding the card's
  total VRAM. With every `VramMiB` at 0 that sum is always 0, so the
  `ErrExceedVramLimit` check silently passes anything — **a fail-open
  validation, not just a display defect.**

Fix: a `MiB` type that decodes any JSON number and rounds to the nearest
whole MiB at the decode boundary, so callers keep working in integer MiB and
no future call site can forget to round. The API DTO (`VgpuProfile.VramMiB`)
stays `uint64`.

---

**Commit 3 — profiles were looked up by PCI address instead of GPU UUID.**

`buildLocalGpuCard` called `getNodeVgpuProfilesMap(hexGpu.PciAddress)`.
`hex_sdk gpu_vgpu_profile_list` resolves the card in
`/etc/cube/cos/gpu/config.json` with `map(select(.id == $id))`, and that `.id`
is the GPU **UUID** — a PCI address matches nothing. The profile catalog
comes from nvidia-smi and is unaffected, so the call returns a full,
plausible-looking list in which the two fields sourced from config.json are
wrong: every profile carries `count: 0` and a nil alias. A wrong answer that
reads as a valid one, with no error anywhere.

Observed on cn13 with one instance genuinely allocated:

```
gpu_vgpu_profile_list GPU-0a5ba9ad-...  -> id 1518 count 1
                                           alias nvidia_..._dc_2b_1518
gpu_vgpu_profile_list 00000001:C8:00.0  -> id 1518 count 0 alias null
```

So `/gpuCards` reported "no profile is configured on this card" while the
card was carved to sriovVgpu with a VM attached to profile 1518.

The update path (`updateLocalGpuCard`) already passes `card.Id`; this listing
call was the only one using the PCI address.
`TestBuildLocalGpuCardVgpuProfiles` asserted that the PCI address was passed
— **the test pinned the defect in place** — and now asserts the UUID.

Also switches `getNodeVgpuProfileCollection` from `CombinedOutput` to
`Output`: `gpu_vgpu_profile_list` writes diagnostics to stderr, and any such
line would be prepended to the JSON and break the unmarshal. This mirrors
`GetNodePgpuAttachedInstance`, which already documents that exact hazard.
stderr is still surfaced on failure via `ExitError`.

**API schema is unchanged across all three commits** — no response field was
added, removed, or retyped, so `api/cube-cos-openapi` needs no bump.

### Test results

1). API docs — not applicable, no schema change (see above). `task
generateApiDocs` still runs clean.

2). Verified on real hardware (cn13, NVIDIA RTX PRO 6000 Blackwell,
2026-07-24 and 2026-07-27), binary hash-traced from the build container all
the way to the node.

**Commit 1 — the point of the change**: with a live vGPU VM running,
`fuser /dev/nvidia0..3` shows **no cube-cos-api process** holding a handle,
and `ls /proc/<pid>/fd | grep nvidia` on the service pid is empty. (The only
holder is `nvidia-vgpu-mgr`, the NVIDIA daemon serving the VM.) With the
pre-change binary restored, the same check immediately showed the service
holding all four devices open continuously. `sriov-manage -e/-d` and full
`gpu_resource_set` carves succeed while this service runs untouched.

Also confirmed for commit 1: `attachedInstances[0].id` equals the VM UUID
from `nvidia-smi vgpu -q` (the PCI-address matching risk originally flagged
as unverified — this is the failure mode that would otherwise look like a
clean, empty response); per-instance FB parse on non-zero data, host
per-instance `FB Used` 128 MiB == `memoryUsage.allocatedMiB` 128, using an
Ubuntu guest with the NVIDIA GRID open kernel module actually loaded.

**Commits 2 and 3** — same machine, same card, same moment, only the binary
changed:

| | before | after |
|---|---|---|
| `degraded` | `true` | `false` |
| sriov profile 1518 | `count 0`, `aliasName null` | `count 1`, `aliasName nvidia_rtx_pro_6000_blackwell_dc_2b_1518` |
| migBacked `vramMiB` (11 profiles) | all `0` | all `23675` |
| `attachedInstances[0].id` | matches VM UUID | matches VM UUID (unchanged) |
| device `vram.allocatedMiB` | 4240 | 4240 (unchanged) |

`23675` is `round(23674.88)`, cross-checked against `nvidia-smi mig -lgip`
reporting 23.12 GiB. `4240` is FB `Reserved` 2288 + `Used` 1952 from a
same-moment `nvidia-smi -q`, confirming the unrelated paths did not regress.
No ERROR/WARN lines in `journalctl -u cube-cos-api` afterwards.

Unit tests: `TestMiBUnmarshalAcceptsFractionalValues` (real 23674.88 fixture,
plus rounding, rejection, and JSON-null cases) and
`TestVgpuProfileCollectionSurvivesFractionalMigVram` (a fractional migBacked
value must not damage the sriov entries decoded alongside it), on top of the
nvidia-smi parser and degrade tests from commit 1. `go test ./internal/...`
passes.


steven-chiu-bigstack (approved):
LGTM! Just noticed that `docs/api/list-local-gpu-cards.md` is still referencing the old implementation (NVML). We can either update it or drop it entirely if it's no longer needed.

Let's ask for @raven-pan's preference on this one. 🙏

PR #623, opened 2026-07-30
fix(integrations): report Cinder storage update times in the node's local offset

raven-pan:
### What type of PR is this?

- Bug fix

<br/>

- Refs bigstack-oss/cubecos#1182
- Handbook write-up: [bigstack-handbook#180](https://github.com/bigstack-oss/bigstack-handbook/pull/180) — **merged 2026-08-01**. It already records this fix in the timestamp-contract Topic and the Cinder known-issue, with the known-issue deliberately left `status: open` until this PR ships. Closing that status is the follow-up once this merges.

<br/>

#### The bugs

**1. The two endpoints disagreed about the same object.** The storage **details** endpoint returned `updateTime` exactly as hex_sdk emitted it (OpenStack UTC). The **list** endpoint converted the same storage's timestamp to the node's local offset.

**2. The list path's fallback used *request* time.** When hex_sdk supplied no update time, the list path substituted `NowRFC3339()`. Two problems, the second being the substantive one:

- it emitted `Z` next to sibling fields carrying the node's local offset, and
- the value **moved on every `GET`** — nothing about the storage had changed.

#### Behaviour, before and after

| case | list — before | details — before | both — after |
|---|---|---|---|
| normal storage | local offset | raw UTC `Z` | local offset |
| no source timestamp | `NowRFC3339()` — UTC, moves every `GET` | empty string | active-firmware time |
| built-in storage | active-firmware time | raw UTC `Z` | active-firmware time |
| unparseable source | left as-is + warn | left as-is | left as-is + warn |

Note the details/no-source-timestamp cell: it returned an **empty string** for a field `GetIntegrationStorageResponse` declares required and non-empty. Now covered by the same placeholder.

#### Changes

```mermaid
flowchart TD
    list["list endpoint<br/>GetStorages"] --> conv["convertStorageTimes"]
    det["details endpoint<br/>GetStorage"] --> convd["convertStorageDetailsTimes<br/>top-level + nested storage"]

    conv --> shared["convertStorageTime<br/><b>single decision point</b>"]
    convd --> shared

    shared --> q1{"isBuiltIn?"}
    q1 -->|yes| fw["base.ActiveFirmwareUpdatedAt<br/>no timestamp of its own<br/>stable across GETs, node offset"]
    q1 -->|no| q2{"updateTime<br/>empty?"}
    q2 -->|yes| fw
    q2 -->|no| q3{"parses as<br/>RFC3339 UTC?"}
    q3 -->|yes| loc["time.LocalRFC3339<br/>node offset"]
    q3 -->|no| raw["pass through<br/>+ log.Warnf"]

    old["<b>was:</b> time.NowRFC3339<br/>request time, UTC Z<br/>moved on every GET"] -.->|replaced by| fw

    style old stroke-dasharray: 5 5
```

**Extract one conversion, call it from both paths.** `convertStorageTime(name, isBuiltIn, updateTime)` holds the whole decision. `convertStorageTimes` (list) and the new `convertStorageDetailsTimes` both delegate to it, so the two endpoints cannot drift apart again.

**The details response carries two update-time fields** — top-level and the one nested under `storage`. Both arrive as raw UTC, so both go through the conversion. `convertStorageDetailsTimes` is nil-tolerant.

**Fallback is now the active-firmware time,** not request time. It is derived once at startup and already carries the node's offset. A built-in storage and one with no source timestamp are the same case — no timestamp of its own — so both branches collapse into one.

**`NowRFC3339` deleted.** No remaining callers, and it emitted UTC in violation of the response-timestamp contract. Follows `ISO8601Z` out (#1166).

#### Compatibility

No schema change. `updateTime` stays required and non-empty on both responses. Only the value's timezone offset and its fallback source change.

Files: `internal/cubecos/integration.go`, `internal/cubecos/integration_test.go`, `internal/definition/v1/time/time.go`.

<br/>

### Test results (optional)

**1). make sure the api docs have been updated**

No API doc change needed — the OpenAPI schema is unchanged (`updateTime` remains a required, non-empty string on both the list and details responses).

<br/>

**2). make sure the api works properly**

`go test ./...` and `go vet ./...` — pass.

New suite in `internal/cubecos/integration_test.go`. Each conversion test runs across three node offsets (`+08:00`, `-05:00`, UTC) so a hardcoded offset cannot pass:

| test | pins |
|---|---|
| `TestConvertStorageTimesReportsTheSourceTimeInTheNodeLocalOffset` | list-path conversion, 3 offsets |
| `TestConvertStorageTimesReportsAStableTimeWhenTheSourceHasNone` | two consecutive calls return the same value — the reported defect |
| `TestConvertStorageTimesKeepsBuiltInStoragesOnTheFirmwareTime` | built-in overrides its source timestamp |
| `TestConvertStorageTimesLeavesAnUnparseableSourceTimeAlone` | bad input passes through instead of becoming empty |
| `TestConvertStorageDetailsTimesConvertsBothUpdateTimeFields` | top-level **and** nested `storage.updateTime`, 3 offsets |
| `TestConvertStorageDetailsTimes…` (matches list / stable fallback / built-in / nil) | the two endpoints agree byte-for-byte; required field never empty; nil does not panic |

**Not verified:** no run against a live cluster. Worth one `GET /v1/integrations/storages` plus `GET /v1/integrations/storages/{name}` on a dev node before merge, confirming both `updateTime` values match and carry the node's offset.

### DIFF
--- a/internal/cubecos/integration.go
+++ b/internal/cubecos/integration.go
@@ -116,6 +116,7 @@ func GetStorage(name string) (*storages.CinderDetails, error) {
 		return nil, err
 	}
 
+	convertStorageDetailsTimes(cinder)
 	return cinder, nil
 }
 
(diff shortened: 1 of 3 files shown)


steven-chiu-bigstack (approved):
LGTM!

PR #625, opened 2026-08-06
fix(gpu): resolve MIG-backed vGPU profiles without a type translation

SekiXu:
### What type of PR is this?

/kind bug

<br/>

Part of bigstack-oss/cubecos#905. Pairs with bigstack-oss/cubecos#1223.

<br/>

### What this PR does?

Removes the MIG-only "vGPU Type ID → GPU Instance Profile ID" reverse lookup.

`nvidia-smi vgpu -q` reports a **vGPU Type ID** per active instance. For
SR-IOV cards that is already what hex records as the profile id, so the lookup
into `hexProfilesMap` was direct. For MIG-backed cards hex used to record the
**GPU Instance Profile ID** instead, so this package ran a second
`nvidia-smi vgpu -s -v` per card to translate one into the other.

As of cubecos#905, `gpu_vgpu_profile_list` reports migBacked profiles per vGPU
type, keyed by vGPU Type ID — the same as the sriov list has always been. The
translation is no longer a translation, so it goes away along with
`GetNvidiaSmiVgpuTypeGpuInstanceProfileIds`, `parseNvidiaSmiVgpuTypeProfileIds`,
`parseNvidiaSmiHexId`, the `nvidiaSmiVgpuTypeProfileIdWindow` constant, the
`MigTypeProfileIds` option field, and the per-card degrade path that fired when
a type had no profile id.

Net effect: one fewer `nvidia-smi` subprocess per MIG-backed card per request,
one fewer parser to keep in sync with nvidia-smi's output format, and
`listVgpuAttachedInstances` treats both vGPU flavours identically.

**Merge ordering.** No longer a question: cubecos#1223 (the hex half of #905)
merged into cubecos `develop` on 2026-08-07 as `ebf47c36`. `gpu_vgpu_profile_list`
there already reports migBacked profiles per vGPU type, keyed by vGPU Type ID —
confirmed against cn13 on 2026-08-10:

```json
{"id":1546,"name":"DC-1-2Q","vramMiB":2048,"vmCountLimit":48,"count":0,"alias":null}
{"id":1549,"name":"DC-1-3Q","vramMiB":3072,"vmCountLimit":32,"count":0,"alias":null}
```

That makes `develop`'s API the mismatched half rather than this branch: its MIG
branch translates a vGPU Type ID into a GPU Instance Profile ID and looks that up
in hex's profile map, where those ids no longer exist, so a MIG-backed card's
`attachedInstances` resolve no `profileAlias`. This PR is what puts the two sides
back in agreement, and nothing has to be sequenced behind it.

`cube-cos-openapi` needs no change — no fields are added or removed, only how
`profileAlias` is resolved internally.

<br/>

### Test results (optional)

No API surface change, so `api/cube-cos-openapi/docs.yaml` is untouched — this
PR only changes how an existing response field is resolved internally. The
`docs.json` derived artifact was still regenerated as part of the build below
(it is gitignored and the local copy was stale), so the Swagger UI served by the
tested binary matches the committed spec.

<br/>

Built x86 in the repo's own jail image (`golang:1.26.5-bookworm`, matching
`build/cube-cos-api-jail.Dockerfile`; the container's own Go was too old for
`go.mod`) and deployed to cn13, the NYCU GPU node — the only node available
that can carve MIG-backed vGPUs at all. Binary identity was confirmed by the
symbols that should have **disappeared**
(`GetNvidiaSmiVgpuTypeGpuInstanceProfileIds`, `parseNvidiaSmiVgpuTypeProfileIds`,
`"no GPU Instance Profile ID found for vgpu type"`) rather than by new strings,
and by checking `GetNvidiaSmiVgpuInstances` was still present. The
previously-deployed binary was verified to still contain the removed symbol, so
the comparison was against a genuinely older build.

With the paired cubecos change carving GPU 0 as `migBackedVgpu` (8 x `DC-1-3Q`,
one 1g.24gb+gfx partition) and a VM booted against the resulting Nova
PCI-passthrough alias:

    GET /api/v1/datacenters/RegionOne/nodes/cn13/gpuCards
    (Bearer token; the API listens on 10.231.0.13:8082, not loopback)

    "resourceType": "migBackedVgpu",
    "status": { "current": "inUse" },
    "migProfileCount": 43,
    "attachedInstances": [{
      "id": "c0822453-59b6-4ae1-8c2f-f7831153c700",
      "name": "mig-1549-vm",
      "profileAlias": "nvidia_rtx_pro_6000_blackwell_dc_1_3q_1549",
      "memoryUsage": { "totalMiB": 3072 }
    }]

- `attachedInstances[0].id` equals the id returned by
  `openstack server create` — the PCI-address string match holds.
- `profileAlias` **resolved**. This is the load-bearing assertion: if the
  removed lookup were still needed, this array would have been silently empty
  rather than erroring.
- `totalMiB` 3072 is `DC-1-3Q`'s framebuffer, and `migProfileCount` 43 is the
  per-type list coming through unchanged.
- Cross-checked against the host: `nvidia-smi vgpu -q -i 0` reported
  `vGPU Type : 1549` for the same VM UUID — the same id hex records, which is
  what makes the translation unnecessary.

`go build ./...`, `go vet`, and `go test -count=1` on
`internal/cubecos/...` and `internal/apis/v1/handlers/nodes/...` all pass
(2026-08-06). Unit tests for the removed parser were deleted with it; no other
test referenced the removed symbols, and a repo-wide grep for all five removed
identifiers returns nothing. `strconv` and `strings` are still used elsewhere in
`gpu_nvidiasmi.go`, so no import became unused.

Re-verified 2026-08-10 in the same x86 jail image, both on this branch alone and
on a local merge with #626 (which touches the same file): `go vet ./internal/...`
exits 0 and `go test -count=1 ./internal/apis/v1/handlers/nodes/...
./internal/cubecos/...` passes on both trees. The merge is clean — #626 changes
the gate that decides *when* profiles are fetched, this PR changes how a fetched
profile is *looked up*, and #626's new tests drive `getNodeVgpuProfilesMap`,
which this PR leaves in the seam list.

Not covered: the SR-IOV path was not re-verified in this pass, as it never went
through the removed lookup — its lookup was already direct and this diff does
not touch it.

Known blind spot, inherited rather than introduced: a vGPU type absent from
hex's profile list is now reported without an alias and without degrading, which
is deliberate (it matches how the SR-IOV path has always behaved). What that
leaves uncovered is a *silently* empty profile map. `gpu_vgpu_profile_list`
redirects nvidia-smi's stderr to `/dev/null` and never checks its exit status
(cubecos `sdk_gpu.sh:580`); its only non-zero return is a missing config.json
(`:553`). Measured on cn13 2026-08-10:

```
$ nvidia-smi vgpu -s -v -i <nonexistent GPU>   exit = 6
$ hex_sdk gpu_vgpu_profile_list <same>         exit = 0
                                      stdout = {"sriov":[],"migBacked":[]}
```

So a failed probe reaches the API as a successful empty fetch, and every
attached instance on that card resolves no alias while the card still reports
`degraded: false`. `develop` behaves the same way today — its MIG branch would
be handed the same empty map — so this is not a regression from this PR, but the
fix belongs in `gpu_vgpu_profile_list`, not here.

### DIFF
--- a/internal/apis/v1/handlers/nodes/gpu.go
+++ b/internal/apis/v1/handlers/nodes/gpu.go
@@ -22,20 +22,19 @@ import (
 // Seams for unit tests: hex_sdk CLI, nvidia-smi, and Openstack are
 // unavailable there.
 var (
-	getNodeGpusMap                 = cubecos.GetNodeGpusMap
-	getNodeVgpuProfilesMap         = cubecos.GetNodeVgpuProfilesMap
-	getNodePgpuAttachedInstance    = cubecos.GetNodePgpuAttachedInstance
-	getNvidiaSmiDevices            = cubecos.GetNvidiaSmiDevices
-	getNvidiaSmiVgpuInstances      = cubecos.GetNvidiaSmiVgpuInstances
-	getNvidiaSmiVgpuTypeProfileIds = cubecos.GetNvidiaSmiVgpuTypeGpuInstanceProfileIds
-	buildInstanceLinks             = buildInstanceLinksViaOpenstack
-	getOpenstackServerNames        = getOpenstackServerNamesViaHelper
-	createConsole                  = createConsoleViaOpenstack
-	getNodeGpuById                 = cubecos.GetNodeGpuById
-	updateNodeGpuCardViaHex        = cubecos.UpdateNodeGpuCard
-	isGpuUpdating                  = isGpuUpdatingViaMongo
-	upsertUpdatingGpuReq           = upsertUpdatingGpuReqViaMongo
-	deleteUpdatingGpuReq           = deleteUpdatingGpuReqViaMongo
+	getNodeGpusMap              = cubecos.GetNodeGpusMap
+	getNodeVgpuProfilesMap      = cubecos.GetNodeVgpuProfilesMap
+	getNodePgpuAttachedInstance = cubecos.GetNodePgpuAttachedInstance
+	getNvidiaSmiDevices         = cubecos.GetNvidiaSmiDevices
+	getNvidiaSmiVgpuInstances   = cubecos.GetNvidiaSmiVgpuInstances
+	buildInstanceLinks          = buildInstanceLinksViaOpenstack
+	getOpenstackServerNames     = getOpenstackServerNamesViaHelper
+	createConsole               = createConsoleViaOpenstack
+	getNodeGpuById              = cubecos.GetNodeGpuById
+	updateNodeGpuCardViaHex     = cubecos.UpdateNodeGpuCard
+	isGpuUpdating               = isGpuUpdatingViaMongo
+	upsertUpdatingGpuReq        = upsertUpdatingGpuReqViaMongo
+	deleteUpdatingGpuReq        = deleteUpdatingGpuReqViaMongo
 )
 
 // buildLocalGpuCardOpts carries data fetched once per listLocalGpuCards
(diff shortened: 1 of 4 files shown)


raven-pan (approved):
LGTM
Thanks for your work!

PR #635, opened 2026-08-17
fix(gpu): report unmeasurable GPU stats as null instead of 0

SekiXu:
### What type of PR is this?

/kind bug

<br/>

bigstack-oss/cubecos#823, and the link half of bigstack-oss/cubecos#826.

<br/>

### What this PR does?

`/gpuCards` answered `0` for numbers that do not exist, so "cannot be measured"
and "idle at 0%" were indistinguishable. Measured on cn13: a MIG-backed vGPU
reports `N/A` for every utilization (NVIDIA provides none once a card is
partitioned) while its framebuffer is fine, and a pgpu is bound to vfio-pci the
moment it is set, so nvidia-smi cannot list the card at all — yet both came back
as confident zeros, `degraded: false`.

Four commits, each independent:

1. **Parser** — fields with no presence guarantee become pointers. A device
   record is only returned when both framebuffer numbers parsed, so those stay
   plain values; every stat on a vGPU instance is a pointer, since an instance is
   reported as soon as it has a VM UUID. Both new fixtures are real cn13 output.
2. **Contract** — `vram`'s three fields, `gpu.utilizationPercent`, and each
   attached instance's `utilizationPercent` / `memoryUsage` are nullable. There is
   **no per-resource-type rule in the handler**: the only rule is that a nil
   source stays nil, so when per-GPU-instance metrics become available (DCGM) the
   numbers appear without touching this layer. `degraded` is deliberately
   untouched — it means enrichment that should have been obtainable was not, while
   these nils are structural facts about the hardware configuration.
3. **`AllTenants`** — the bulk server prefetch that replaced a per-instance
   `GetServer` dropped it, which narrowed the listing to the API's own project.
   Fetching one server by id is not project-scoped for an admin token, so the
   behaviour changed silently: a vGPU VM owned by any other project came back
   nameless with only a warning. Invisible on a lab node where every VM is in
   `admin`.
4. **Grafana link** — pinning `var-UUID` alone left the dashboard's
   `TID → TENANT → HOSTNAME → UUID` chain to resolve on load, so the page charted
   the requested VM under the tenant's alphabetically first VM name (observed on
   cn13 naming an already deleted VM above a running one's data). All three are
   pinned now, and a panel is opened explicitly because the vGPU row ships
   collapsed. `kiosk=tv` is dropped — TV mode was removed in Grafana 11 and the
   product ships 12.3.2.

⚠️ **Cross-team contract**: instance dashboard panel **37** (Memory Usage) is now
pinned by these links, like the device dashboard's panels 50/51. Renumbering or
reordering that row breaks them. Memory is the panel chosen because it renders for
every vGPU flavour; GPU utilization is empty for a MIG-backed vGPU by design.

<br/>

### Test results (optional)

Companion PR bigstack-oss/cube-cos-openapi#107 — **merge that one first**, then
this PR's submodule pointer is re-bumped to the merged sha.

`go vet` clean and `go test -count=1` green in the x86 jail container (4 new
tests: MIG device parse, MIG-backed instance parse, pgpu nil stats, link
variables + escaping). Deployed on cn13, the only node with vGPU:

| card | type | `vram` | `gpu.util` | attached instance | degraded |
| --- | --- | --- | --- | --- | --- |
| 42:00.0 | sriovVgpu | 14128 / 97887 / **0** | **0** | `k8s-gpu-sriov` util **0**, mem 512/12288 | false |
| 8C:00.0 | migBackedVgpu | 13949 / 97887 / **null** | **null** | `k8s-gpu-mig` util **null**, mem 784/12288 | false |
| 04:00.0 | migBackedVgpu | 4317 / 97887 / **null** | **null** | `test-mig-vgpu` util **null**, mem **144/2048** | false |
| C8:00.0 | pgpu | **null / null / null** | **null** | none (idle) | false |

The sriovVgpu `0` is a real zero (idle card), so this is not a regression, and
`04:00.0`'s 144/2048 matches `nvidia-smi vgpu -q` exactly.

Not verified on hardware: a pgpu **with a VM attached** — that card is idle right
now, so only the unit test covers it.

Known limitation worth a reviewer's opinion: a pgpu-only node skips the server
prefetch, so such a link carries an empty `var-TID` and the header can still be
off. The UI design disables the Grafana link on a pgpu row anyway (no series
exists for that type), so I left it.

> **Merge this one first.** cube-cos-api#636 (cubecos#824/#825, per-card history
> links) is stacked on top of this branch: integrating the two conflicted in four
> files, so rather than leave that for whoever merged second, #636 now carries these
> four commits and the conflicts are already resolved there. Once this lands, #636
> becomes a single-commit fast-forward.

> **Correction — the frontend already consumes these fields.** I claimed the
> opposite when opening this PR, and the claim is wrong. `cube-cos-ui` renders them
> today, and the `!row.vram` checks are object-level rather than field-level:
>
> | site | with this PR |
> | --- | --- |
> | `NodeGpuResources.tsx:122-124` | `toReadableUsedSize({used: number})` gets `number \| null` → **build fails** |
> | `GpuDetails.tsx:42-46` | same, for the instance's `memoryUsage` → **build fails** |
> | `NodeGpuResources.tsx:134` | `` `${row.vram.utilizationPercent}%` `` → prints **`null%`** |
> | `NodeGpuResources.tsx:140` | `` `${row.gpu.utilizationPercent}%` `` → prints **`null%`** |
> | `GpuDetails.tsx:118-120` | `` `${utilizationPercent} %` `` → prints **`null %`** |
>
> So this PR needs a paired UI change: two of those fail to compile once the SDK
> regenerates, three would ship `null%` silently. The nullable contract is still the
> right call -- and the two build failures are the compile-time check that argues for
> `null` over `""` -- but it is not free, it needs the frontend in the same release.
>
> **Added after review** (`b3ab96bd`): an attached instance's `links.grafana` is now
> nullable too, and is omitted when the dashboard variables cannot be pinned -- see
> the review thread. Depends on cube-cos-openapi#109.

### DIFF
--- a/internal/apis/v1/handlers/grafana/links.go
+++ b/internal/apis/v1/handlers/grafana/links.go
@@ -2,6 +2,7 @@ package grafana
 
 import (
 	"fmt"
+	"net/url"
 
 	"github.com/bigstack-oss/cube-cos-api/internal/definition/v1/base"
 	"github.com/gin-gonic/gin"
@@ -15,14 +16,49 @@ func genHostLink(c *gin.Context) string {
 	)
 }
 
-// InstanceDashboardLink is the Grafana instance dashboard URL for a VM. It is
-// exported for other handlers (e.g. GPU attached instances) that report links
-// to the same dashboard.
+// InstanceDashboardLink is the Grafana instance dashboard URL for a VM, as
+// served by the generic /grafana/instances/{instanceId} endpoint. That endpoint
+// knows nothing but the id, so the link cannot pin the dashboard's tenant and
+// hostname variables; callers that do know them should use
+// InstanceVgpuDashboardLink instead.
 func InstanceDashboardLink(instanceId string) string {
 	return fmt.Sprintf(
 		"https://%s/grafana/d/PVW6vU7Wz/instance?refresh=5m&kiosk=tv&orgId=1&var-UUID=%s",
 		base.DataCenterVip,
 		instanceId,
+	)
+}
+
+// instanceVgpuMemoryPanelId is the Memory Usage panel in the instance
+// dashboard's vGPU row (dashboard UID PVW6vU7Wz). Pinned here as a deep-link
+// target, so renumbering or reordering that row's panels in instance.json
+// breaks these links -- same cross-team contract as the device dashboard's
+// panels 50 and 51.
+const instanceVgpuMemoryPanelId = 37
+
+// InstanceVgpuDashboardLink is the deep link to one VM's vGPU usage.
+//
+// Pinning var-UUID alone is not enough. UUID is the tail of a
+// TID -> TENANT -> HOSTNAME -> UUID chain of query variables that all refresh on
+// dashboard load, so leaving the first two unset lets them resolve to whatever
+// their query returns first -- in practice the tenant's alphabetically first VM,
+// which then labels the page with a different VM than the one being charted.
+// var-TID is matched against the metrics' tenant_id tag and collapses TENANT to
+// the owning project; var-HOSTNAME pins the picker to this VM.
+//
+// viewPanel is required because the vGPU row ships collapsed: a link to the
+// dashboard alone lands on a page with no GPU chart visible. Memory is the panel
+// to open because it renders for every vGPU flavour, while GPU utilization is
+// empty for a MIG-backed vGPU by design.
+func InstanceVgpuDashboardLink(instanceId, tenantId, vmName string) string {
+	return fmt.Sprintf(
+		"https://%s/grafana/d/PVW6vU7Wz/instance?orgId=1&refresh=5m&var-TID=%s&var-HOSTNAME=%s&var-UUID=%s&from=now-3h&to=now&viewPanel=%d",
+		base.DataCenterVip,
+		url.QueryEscape(tenantId),
+		// A VM name is user-supplied and may contain spaces or '&'.
+		url.QueryEscape(vmName),
+		url.QueryEscape(instanceId),
+		instanceVgpuMemoryPanelId,
 	)
 }
 
--- a/internal/apis/v1/handlers/grafana/links_test.go
+++ b/internal/apis/v1/handlers/grafana/links_test.go
@@ -13,3 +13,26 @@ func TestInstanceDashboardLink(t *testing.T) {
 	require.Contains(t, link, "/grafana/d/PVW6vU7Wz/instance")
 	require.True(t, strings.HasSuffix(link, "var-UUID=vm-1"))
 }
+
+func TestInstanceVgpuDashboardLink(t *testing.T) {
+	link := InstanceVgpuDashboardLink("vm-1", "proj-1", "instance-1")
+
+	require.Contains(t, link, "/grafana/d/PVW6vU7Wz/instance")
+	// Pinning all three is what keeps the page from labelling this VM's chart
+	// with the tenant's alphabetically first VM.
+	require.Contains(t, link, "var-TID=proj-1")
+	require.Contains(t, link, "var-HOSTNAME=instance-1")
+	require.Contains(t, link, "var-UUID=vm-1")
+	// The vGPU row ships collapsed, so the link has to open a panel directly.
+	require.Contains(t, link, "viewPanel=37")
+	// kiosk=tv was removed in Grafana 11; it must not be carried over.
+	require.NotContains(t, link, "kiosk")
+}
+
+// A VM name is user-supplied: unescaped, a space or '&' would truncate or
+// corrupt the variable that follows it.
+func TestInstanceVgpuDashboardLinkEscapesValues(t *testing.T) {
+	link := InstanceVgpuDashboardLink("vm-1", "proj-1", "my vm&x")
+
+	require.Contains(t, link, "var-HOSTNAME=my+vm%26x")
+}
--- a/internal/apis/v1/handlers/nodes/gpu.go
+++ b/internal/apis/v1/handlers/nodes/gpu.go
@@ -671,7 +671,12 @@ func createConsoleViaOpenstack(vmId string) (*remoteconsoles.RemoteConsole, erro
 // getOpenstackServerNamesViaHelper returns a map of server id to name in a single
 // Openstack round trip, used to resolve vGPU attached-instance names in bulk.
 func getOpenstackServerNamesViaHelper() (map[string]string, error) {
-	serverList, err := openstack.GetGlobalHelper().ListServers(servers.ListOpts{})
+	// AllTenants is required: the API authenticates as a single project (admin),
+	// so a plain list returns only that project's servers and a vGPU VM owned by
+	// any other project would come back nameless. The per-instance GetServer this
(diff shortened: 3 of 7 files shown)


raven-pan:
Nice work on the parser layer — the indent-based section tracking makes the MIG
device block safe regardless of where nvidia-smi puts it, which is stronger than
the fixture comment claims.

One question on the known limitation you flagged for a reviewer's opinion.

## The UI does not disable the link on a pgpu row

`cube-cos-ui` guards only the alias column on `resourceType`
(`GpuDetails.tsx:110`). The action column that carries the Grafana link has no
such guard (`GpuDetails.tsx:129-132`), and `renderAttachedInstanceActions`
renders the hyperlink unconditionally (`GpuDetails.tsx:61`).
`FullViewInstanceTable.tsx:68` does the same. So a pgpu row ships a live,
clickable link today.

## The result is worse than a wrong header

`cubecos/core/grafana/provisioning/dashboards/instance.json` declares `TID` as a
textbox and `TENANT` as:

```
SHOW TAG VALUES FROM "vm.host_alive_status" WITH KEY = "tenant_name" WHERE tenant_id =~ /$TID/
```

An empty `$TID` makes that regex match every tenant. On dashboard load:

1. `TENANT` resets to the first tenant.
2. `HOSTNAME` options collapse to that tenant's VMs, so the pinned
   `var-HOSTNAME` is no longer an option and resets.
3. `var-UUID` resets for the same reason.

The page then charts a different VM, not the requested VM under a wrong label.
That is the failure mode this PR removes everywhere else.

## Proposal: report no link when the variables cannot be pinned

This follows the rule you already set for the stats — a nil source stays nil. A
link that cannot be built correctly is the same kind of absence as a number that
cannot be measured.

```go
type InstanceLinks struct {
	// Grafana is nil when the dashboard variables cannot be pinned. Without the
	// tenant id, TENANT/HOSTNAME/UUID all reset on load and the page charts a
	// different VM. A missing link is honest; a link to the wrong VM is not.
	Grafana *string `json:"grafana"`
}

func buildInstanceLinksViaOpenstack(opts instanceLinkOpts) gpu.InstanceLinks {
	if opts.TenantId == "" || opts.VmName == "" {
		return gpu.InstanceLinks{}
	}
	link := grafana.InstanceVgpuDashboardLink(opts.InstanceId, opts.TenantId, opts.VmName)
	return gpu.InstanceLinks{Grafana: &link}
}
```

The UI guard is one line and accepts `null`, `""`, and `undefined` alike:

```tsx
{row.links.grafana && (
  <CosHyperlink href={row.links.grafana} target="_blank">Grafana</CosHyperlink>
)}
```

## Why `null` and not `""`

`docs.yaml:15724` types `grafana` as a plain `string` with no `format` or
`pattern`, so `""` is already schema-valid and needs no spec change. The
difference is enforcement:

| | `""` | `null` |
| --- | --- | --- |
| openapi change | none | `nullable: true` |
| SDK type | `string` | `string \| null` |
| A call site that forgets the guard | compiles, fails silently | fails to compile |

There are two call sites today. With `""`, a third one added later brings the bug
back and nothing catches it. This PR already turns six fields nullable and
already needs an openapi and a UI change, so one more nullable field adds no new
coordination and buys a compiler check.

## It also covers a second path

`gpu.go:625`: a vGPU instance absent from the prefetched server map falls through
with an empty `TenantId` and an empty `Name`. That is a vGPU row, so the pgpu
argument does not reach it. The nil-link rule handles both paths in one place.

Side effect: a pgpu-only node loses its Grafana links. That matches your own
reasoning — no series exists for a pgpu — and closes the gap where the UI was
expected to hide the link but does not.

Does this seem reasonable to you? If there is a reason to keep the link on a pgpu row, or if you prefer `""` to keep
the spec change smaller, I am happy either way. Your call.

raven-pan (approved):

SekiXu:
Adopted in `b3ab96bd`, and `cube-cos-openapi#109` carries the spec change. Two
corrections and one finding that makes your proposal more necessary than either of
us thought.

## Your UI reading is right, and it is worse than the link

`utilizationPercent` and `memoryUsage` are already rendered today, unguarded — the
`!row.vram` checks are object-level, not field-level:

| site | with this PR |
| --- | --- |
| `NodeGpuResources.tsx:122-124` | `toReadableUsedSize({used: number})` gets `number \| null` → **build fails** |
| `GpuDetails.tsx:42-46` | same, for the instance's `memoryUsage` → **build fails** |
| `NodeGpuResources.tsx:134` | `` `${row.vram.utilizationPercent}%` `` → prints **`null%`** |
| `NodeGpuResources.tsx:140` | `` `${row.gpu.utilizationPercent}%` `` → prints **`null%`** |
| `GpuDetails.tsx:118-120` | `` `${utilizationPercent} %` `` → prints **`null %`** |

So my "the frontend has not consumed these fields yet, changing now is free" claim
was wrong, and it is the premise I put in the issue, this PR and the design doc. It
needs a paired UI change either way: two sites fail to compile, three would ship
`null%` silently. I am correcting that everywhere and handing the frontend this list
rather than the reassurance.

That also settles your `null` vs `""` question empirically, in your favour: the two
sites that break the build are exactly the compiler check you were arguing for, and
`href` on `CosHyperlink` is typed `string`, so a null link fails to compile too. `""`
buys none of that.

## The failure mode is milder than "charts a different VM"

Measured on cn13, Grafana 12.3.2 keeps a URL-supplied variable value even when the
refreshed option list excludes it:

- `var-UUID=00000000-0000-0000-0000-000000000000` with no `var-TID`/`var-HOSTNAME`:
  the `Id` picker *displays that nonexistent UUID* and the panel says `No data` — it
  does not fall back to another VM.
- `var-TID=` (empty) with `var-HOSTNAME` and `var-UUID` pinned: the chart is the
  requested VM's, all three pickers correct.

So what an unpinnable link produces here is the right chart under a possibly wrong
label, not the wrong chart. Caveat on my own evidence: cn13 has a single tenant with
VMs, so "TENANT reset to the first tenant" and "TENANT happens to be correct" are
indistinguishable there. Your cascade needs HOSTNAME/UUID to be *dropped* when
absent from the refreshed options, which the first measurement contradicts directly.

None of that changes the outcome: a pgpu has no `gpu.vm` series, and the link carries
`viewPanel=37`, so it opens straight onto an empty panel by construction. Not
shipping it is right — I just did not want to justify it with a mechanism I could not
reproduce.

## Where I did not extend the rule

The card's own `links.workloadHistory` / `links.vramHistory` (#636) stay non-null.
The distinction is wrongly-pinned versus correctly-pinned-but-empty:

| | instance link | card link |
| --- | --- | --- |
| missing input | pgpu has no resolvable tenant | nothing — hostname and pciAddress are always there |
| what it points at | variables may resolve elsewhere, and the panel is empty anyway | the right card, whose chart may be empty |

A pgpu card's history links open empty panels 50/51, which is the same honest answer
as the `—` its stats now render. If you would rather they were null too, say so and I
will follow the same rule there.

## Also in this push

`getGpuTotalVramMiB` stays untouched: a device record is only returned when both
framebuffer numbers parsed (`parseNvidiaSmiDevices`), so that path has no new nil
branch and the VRAM-limit validation is unchanged.

PR #636, opened 2026-08-18
feat(gpu): report per-card GPU history links on the card listing

SekiXu:
### What type of PR is this?

/kind feature

<br/>

bigstack-oss/cubecos#824 and #825 (the API half).

<br/>

### What this PR does?

The UI offers "View Workload History" and "View VRAM History" from each GPU row's
menu, but the links available so far were built per node: every row on the same
node produced the same URL and the chart answered for the whole node. Worse, from
a MIG card's row you would see the other cards' lines and none for the card you
actually asked about.

Report the two links **inline with each card** in
`GET .../nodes/{nodeName}/gpuCards`:

```json
"links": {
  "workloadHistory": "https://<vip>/grafana/d/i-device/device?orgId=1&var-GPU_HOST=cn13&var-GPU_PCIID=00000001%3A04%3A00.0&from=now-3h&to=now&viewPanel=50",
  "vramHistory":     "…&viewPanel=51"
}
```

Inline rather than behind their own endpoint, for the same reason the attached
instances carry their Grafana link inline: the UI already fetches this response to
draw the table, and both values are pure string building from data in hand (node
name + the card's PCI address), so a per-row action costs no extra request.

The card is named by the new hidden `$GPU_PCIID` dashboard variable
(bigstack-oss/cubecos#1287). Its values are `gpu.host`'s `pciid` tag, which is
byte-identical to the card's `pciAddress` — verified on cn13 across all four cards
— so nothing has to be translated.

Two details worth knowing:

- **Empty pciAddress omits the variable** rather than sending it empty. The
  variable defaults to `All`; an empty value would match no card at all. That is
  what the two node-scoped `/grafana/devices/{hostname}/…` endpoints keep doing —
  they are handed only a hostname, so they cannot name a card. They are unchanged.
- **A card with no series still gets a link.** An empty chart is the honest answer
  for a passed-through GPU or for utilization on a MIG-enabled card, and it matches
  the null stats the same card reports, so nothing needs disabling in the UI.

⚠️ **Cross-team contract**: panel ids **50** (GPU Utilization) and **51** (VRAM)
in `device.json` are pinned by these links, and so is the variable name
`var-GPU_PCIID`. Renumbering or reordering that row breaks them.

<br/>

### Test results (optional)

Companion PR bigstack-oss/cube-cos-openapi#108 — **merge that one first**, then
this PR's submodule pointer is re-bumped to the merged sha.

`go vet` clean and `go test -count=1` green in the x86 jail container; build
produces an x86-64 binary with the regenerated spec embedded. New tests: per-card
link pins host + escaped PCI address + the right panel, and an empty PCI address
drops the variable instead of emptying it; the listing test asserts the links on
both a visible and a passthrough card.

> Depends on cubecos#1287 for the dashboard variable. Until that lands, a link
> carrying `var-GPU_PCIID` is simply ignored by the dashboard — it does not break,
> it just still shows every card.
>
> **Stacked on cube-cos-api#635 — the conflicts are already resolved.**
> An earlier version of this description claimed the two would merge cleanly in
> either order. They do not: integrating them locally conflicted in four files, so
> this branch is now rebased on top of #635 and carries its four commits as well.
> Nobody has to resolve anything at merge time.
>
> What conflicted, for the record:
>
> | file | conflict |
> | --- | --- |
> | `handlers/grafana/links_test.go` | both append tests at end of file |
> | `definition/v1/gpu/gpu.go` | one turns fields into pointers, the other adds `GpuCardLinks`, adjacent lines |
> | `handlers/nodes/gpu.go` | seam block and the new function land next to each other — and this change referenced `getOpenstackServerNames`, which #635 renames to `getOpenstackServers` |
> | `handlers/nodes/gpu_test.go` | seam save/restore, two places |
>
> All additive except the rename, which is why resolving it mechanically would have
> left code that does not compile.
>
> **Merge order: #635 first.** Once it lands, `develop` is this branch's parent and
> what remains here is a single commit — a plain fast-forward. Reviewing only this
> change is `git diff fix/823-gpu-usage-unavailable-vs-zero..fix/824-825-per-gpu-history-link`
> (5 files, +105/-12).
>
> The integrated result is what was verified: `go vet` clean, `go test -count=1`
> green, and both features confirmed together on cn13 (per-card links narrow panel
> 51 from six series to two, and the null stats still hold for MIG-backed and
> pgpu cards).

### DIFF
--- a/internal/apis/v1/handlers/grafana/links.go
+++ b/internal/apis/v1/handlers/grafana/links.go
@@ -101,21 +101,56 @@ func genStoragesLink() string {
 	)
 }
 
-// panel 50 = GPU Utilization on the device dashboard (UID i-device).
-// Filtered by the hidden $GPU_HOST variable (gpu.host's `host` tag), NOT $HOST.
-func genGpuUtilizationHistoryLink(c *gin.Context) string {
-	return fmt.Sprintf(
-		"https://%s/grafana/d/i-device/device?orgId=1&var-GPU_HOST=%s&from=now-3h&to=now&viewPanel=50",
+// The device dashboard's GPU panel ids, pinned here as deep-link targets:
+// renumbering or reordering them in device.json breaks these links.
+const (
+	deviceGpuUtilizationPanelId = 50
+	deviceGpuVramPanelId        = 51
+)
+
+// deviceGpuHistoryLink builds a device-dashboard deep link (UID i-device) to one
+// GPU history panel.
+//
+// $GPU_HOST is gpu.host's `host` tag, NOT $HOST (which comes from ipmi_sensor and
+// can differ). pciAddress is optional and scopes the chart to a single card via
+// the hidden $GPU_PCIID variable, whose values are gpu.host's `pciid` tag --
+// byte-identical to the pciAddress the GPU listing reports for the same card, so
+// no translation is needed. Passing an empty pciAddress omits the variable
+// entirely rather than sending it empty: it defaults to All, and an empty value
+// would match no card at all.
+func deviceGpuHistoryLink(hostname, pciAddress string, panelId int) string {
+	link := fmt.Sprintf(
+		"https://%s/grafana/d/i-device/device?orgId=1&var-GPU_HOST=%s",
 		base.DataCenterVip,
-		c.Param("hostname"),
+		url.QueryEscape(hostname),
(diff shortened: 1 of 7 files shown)


raven-pan (approved):
LGTM

PR #637, opened 2026-08-19
feat(gpu): report workload and VRAM history links on a GPU-attached instance

SekiXu:
Closes part of bigstack-oss/cubecos#823. Spec PR: bigstack-oss/cube-cos-openapi#110.

**Merge order:** openapi#110 first, then I re-bump the submodule pointer here to its merged sha, then this one. Merging this first would pin a commit that only exists on a branch.

## What changes

`attachedInstances[].links` on `GET /nodes/{nodeName}/gpuCards`:

```diff
 links:
-  grafana: <instance dashboard, viewPanel=37>
+  workloadHistory: <instance dashboard, viewPanel=36>   # GPU Usage
+  vramHistory:     <instance dashboard, viewPanel=37>   # Memory Usage
```

The single link was hardwired to Memory Usage, so a VM's **workload** history was reachable only by opening the instance dashboard's collapsed vGPU row by hand. The new pair is named to match the `workloadHistory` / `vramHistory` each GPU card already reports, so the card table and the attached-instance table speak one contract.

Both stay inline with the listing (pure string building from data already in hand — a per-row action costs no extra request). The console stays on its own endpoint for the opposite reason: minting one is a Nova write.

## Behaviour rules, unchanged in spirit

| case | workloadHistory | vramHistory | why |
| --- | --- | --- | --- |
| SR-IOV vGPU | link | link | both panels have series |
| MIG-backed vGPU | link (**empty panel**) | link | hardware reports no utilization; see below |
| tenant or VM name unresolvable (pgpu-only node, instance missing from the prefetch) | `null` | `null` | the dashboard's variables cannot be pinned, so the page would label this VM's chart with another VM |

The two are omitted **together** — a half-pinned pair mislabels just as badly as one.

## The one decision worth a second opinion

A MIG-backed vGPU has **no utilization series at all**, so its `workloadHistory` opens an empty panel. Measured on cn13 today (all three resource types carved simultaneously):

```
influx -database telegraf -execute 'SELECT last(util_gpu) AS util, last(mem_used) AS mem FROM "gpu.vm" WHERE time > now()-30m GROUP BY vm_uuid, "name"'

k8s-gpu-mig    (DC-1-12Q, MIG-backed)  util: <field absent>   mem: 784
test-mig-vgpu  (DC-1-2Q,  MIG-backed)  util: <field absent>   mem: 144
k8s-gpu-sriov  (DC-12Q,   SR-IOV)      util: 0                mem: 512
```

`util_gpu` is **absent**, not 0 — consistent with `nvidia-smi vgpu -q` reporting `Utilization → N/A` for a MIG-backed vGPU while `FB Memory Usage` is fine, and with `utilizationPercent: null` that this same response already reports for those two instances.

I kept the link and let the panel be empty, matching the merged card-level stance ("a card with no series … still gets a link; the chart is simply empty"). The alternative — `workloadHistory: null` for MIG-backed rows so the UI can disable the button — is defensible too and is a one-line change in each PR. @raven-pan your call; I'll switch both before merge if you prefer it.

## ⚠️ Frontend must land in the same release

This renames a field. `cube-cos-ui` reads `links.grafana` on an attached instance in `packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeGpuResources/GpuDetails.tsx` (Action column) — that fails to compile against the regenerated SDK, which is the intended outcome rather than silently reading `undefined`. The same row gains a second button; Console is untouched (still `GET …/gpuCards/instances/{instanceId}/console`, minted on demand).

## Submodule pointer

The bump here also catches api up to the **four** GPU doc commits already merged in cube-cos-openapi that `develop`'s pointer predates (`fab3c3b`, `ad46f27`, `c865681`, plus `058c464`). Until now the binary's embedded `docs.json` — hence the live Swagger — described neither the nullable GPU stats nor the per-card history links that are already shipping.

## Verification

Built and tested in the x86 jail container (never on the arm64 Mac):

```
yq -o=json -I=4 api/cube-cos-openapi/docs.yaml > api/docs.json     # 102 paths
CGO_ENABLED=1 GOWORK=off GOOS=linux GOARCH=amd64 go build -o bin/cube-cos-api cmd/main.go   # ok
CGO_ENABLED=1 GOWORK=off go test ./...                            # all packages ok, no FAIL
CGO_ENABLED=1 GOWORK=off go vet ./internal/apis/v1/handlers/...    # clean
```

Unit tests cover: both links pin all three dashboard variables and open panel 36 / 37 respectively; a user-supplied VM name stays escaped; both links absent when tenant or name is missing.

**Not yet verified live** — the binary is not deployed to cn13 (the only vGPU node), so the JSON above has not been curled from a running build. The pre-change response was curled there today, which is where the panel/series evidence comes from.

### DIFF
--- a/internal/apis/v1/handlers/grafana/links.go
+++ b/internal/apis/v1/handlers/grafana/links.go
@@ -19,8 +19,8 @@ func genHostLink(c *gin.Context) string {
 // InstanceDashboardLink is the Grafana instance dashboard URL for a VM, as
 // served by the generic /grafana/instances/{instanceId} endpoint. That endpoint
 // knows nothing but the id, so the link cannot pin the dashboard's tenant and
-// hostname variables; callers that do know them should use
-// InstanceVgpuDashboardLink instead.
+// hostname variables; callers that do know them should use the per-VM history
+// links below instead.
 func InstanceDashboardLink(instanceId string) string {
 	return fmt.Sprintf(
 		"https://%s/grafana/d/PVW6vU7Wz/instance?refresh=5m&kiosk=tv&orgId=1&var-UUID=%s",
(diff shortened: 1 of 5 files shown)


raven-pan (approved):
LGTM

PR #639, opened 2026-08-21
fix(gpu): unblock pgpu -> vGPU switches + stop cutting off hex mid-carve

### LINKED ISSUE #638: [Bug]: switching a pgpu to either vGPU type always 500s on nvidia-smi, before hex is called
## Description

`PUT .../nodes/{nodeName}/gpuCards/{gpuId}` cannot switch a card away from `pgpu` to either vGPU type when profiles are supplied. It always fails with a 500 **before `hex_config` is ever called**, so the card is left untouched:

```json
{"code":500,"msg":"nvidiasmi: device GPU-0a5ba9ad-a575-72c7-d51d-3e726232cfd1 not reported by nvidia-smi","status":"internal server error"}
```

Reproduced on cn13 (RTX PRO 6000 Blackwell), 100%, for **both** target types:

```
pgpu -> sriovVgpu      : 500 nvidiasmi: device GPU-0a5ba9ad-… not reported by nvidia-smi
pgpu -> migBackedVgpu  : 500 nvidiasmi: device GPU-0a5ba9ad-… not reported by nvidia-smi
```

### Root cause — a timing problem, not a data problem

`updateLocalGpuCard` (`internal/apis/v1/handlers/nodes/gpu.go`) runs its pre-flight validation **before** calling `hex_config gpu_resource_set`, and part of that validation reads the card's total VRAM from `nvidia-smi`. But a `pgpu` is bound to `vfio-pci` and is therefore **invisible to nvidia-smi** — hex only hands the card back to the nvidia driver *inside* `gpu_resource_set`, i.e. after this validation has already run.

Two independent defects stack up:

1. **The query is wider than its use.** `totalVramMiB` is only ever consulted for `migBackedVgpu` (`gpu.go:500`; SR-IOV profiles are fixed partitions bounded by the profile-count limit instead), but it was gathered for `isVgpuType()`, which includes `sriovVgpu`. An SR-IOV request was rejected on a value it would never have looked at.
2. **The check runs where it cannot possibly succeed.** `config_gpu.cpp`'s `gpu_resource_set` (L1122-1160) already does `gpu_unbind_vfio_pci` → `ValidateVgpuProfiles` → re-bind to `vfio-pci` if the checks fail. Its `ValidateVgpuProfiles` (L981-991) enforces the **identical** VRAM rule, at the correct point in the sequence, and **fails closed**. The API-side copy is a shadow of it.

Note the read path already tolerates this: `gpu.go:261` *degrades* a card nvidia-smi cannot see rather than failing. Only the write path turned it into a 500.

## Second bug found while verifying this: a 30s timeout that reports success as failure

`UpdateNodeGpuCard` (`internal/cubecos/nodes.go:419`) gave `hex_config gpu_resource_set` a 30s context. A resource switch mutates hardware — it releases the card from vfio-pci, carves it (48 VFs for SR-IOV; MIG mode plus GPU instances for MIG-backed), rewrites `/etc/nova/nova.d/gpu.conf` and restarts the nova services. Measured on cn13 it takes **25–60s**.

So `pgpu -> sriovVgpu` returned `{"code":500,"msg":"signal: killed"}` after 37.7s while `hex_config` **carried on and completed** (journal shows it restarting nova-conductor/scheduler/api/compute afterwards); `config.json`, `sriov_numvfs=48` and `driver=nvidia` all landed correctly. **The caller saw a 500 for an operation that succeeded.** The `migBackedVgpu` case returned 200 at 30.4s purely by luck.

Worse than a slow request: `exec.CommandContext` **SIGKILLs** hex_config on expiry — killing a process partway through repartitioning a GPU, after which the hardware, `config.json` and nova's config are free to disagree.

## Solution

1. Narrow the total-VRAM query to `migBackedVgpu`, and make it optional: when nvidia-smi cannot report the card, skip the VRAM check rather than failing. Nothing is bypassed — hex enforces the same rule after the unbind, fails closed, and rolls the card back to vfio-pci if it trips.
2. Raise the switch's context timeout to 180s (matching the existing precedent in `firmware.go:84`). The read-only `hex_sdk` queries keep their 30s.

## Verification (cn13, RTX PRO 6000 Blackwell)

Test card `GPU-0a5ba9ad-…` (`00000001:C8:00.0`), `pgpu` / `idle` / `allocation.current=0`. The node's other three cards were `inUse` and untouched throughout.

| # | operation | path | elapsed | result |
|---|---|---|---|---|
| 1 | pgpu → sriovVgpu | via haproxy | 33.4s | **200 OK** (was `500 signal: killed` on the 30s build) |
| 2 | sriovVgpu → pgpu | via haproxy | 26.4s | 200 OK |
| 3 | pgpu → migBackedVgpu | via haproxy | 60.05s | cut off by haproxy — see below |
| 4 | pgpu → migBackedVgpu | direct to :8082 | 32.3s | 200 OK |

Card state after #1: `sriov_numvfs=48`, `driver=nvidia`, VF `0001:c8:00.2` carrying `current_vgpu_type=1518`, and `/etc/nova/nova.d/gpu.conf` holding the alias `nvidia_rtx_pro_6000_blackwell_dc_2b_1518`.
Card state after #4: `mig.mode.current=Enabled`, GI created (`MIG 1g.24gb+gfx`, profile 47), `driver=nvidia`.

### The skipped check is not a hole

Card total = **97887 MiB**; profile 1546 = 2048 MiB with `vmCountLimit=48`. `count=48` → 98304 MiB therefore exceeds VRAM **without** exceeding the count limit — a clean test vector.

| case | result |
|---|---|
| unknown profile id 99999 (card visible) | `400 vgpu profile 99999 not found` |
| 98304 > 97887, card **visible** | `409 requested vram 98304 MiB exceeds device total 97887 MiB` — caught by the API |
| same request, card **invisible** (pgpu) | `500 exit status 1`; hex journal: `gpu_resource_set: requested 98304 MiB of MIG-backed vGPU memory exceeds the 97887 MiB available` — **caught by hex, and the card was correctly rolled back to pgpu + vfio-pci** |

Unit tests: `go test -race ./internal/apis/v1/handlers/nodes/` green on both arm64 and the x86 jail container. Four cases added — the SR-IOV path asserts nvidia-smi is **not** queried; the MIG path proceeds when the card is invisible and when nvidia-smi is unavailable; and the limit still bites when nvidia-smi does report the card.

## Known remaining issue — not fixed here

The 180s budget makes the false failure rarer, **not impossible**. The API is fronted by haproxy with `timeout server 1m`, and nginx's `/cos-api/` location takes the 60s `proxy_read_timeout` default. Measured switch durations (25.3s – 60.05s) sit **right against that 60s ceiling**, and case #3 above crossed it: the client got a gateway timeout with no body while the switch completed normally. Compare #3 with #4 — same operation, 60.05s and cut off through the proxy, 32.3s and a 200 direct to `:8082`.

What 180s buys is that hex_config is no longer SIGKILLed midway through repartitioning a GPU. Making the *response* honest requires the endpoint to go asynchronous (return 202 and let the caller poll `Status.IsProcessing`, which the `ReqGpuCollection` record already drives). That is a breaking API change needing UI coordination, so it is deliberately out of scope here and should get its own ticket.

SekiXu:
Closes #638. No OpenAPI change: status codes and response schema are untouched, so there is no companion `cube-cos-openapi` PR and no merge-order dance.

## What changes

Two independent fixes, one per commit.

### 1. `pgpu` → vGPU no longer 500s on nvidia-smi

`updateLocalGpuCard` read the card's total VRAM from `nvidia-smi` as a pre-flight check. A `pgpu` is bound to `vfio-pci` and is invisible to nvidia-smi — hex only hands it back to the nvidia driver *inside* `gpu_resource_set`, i.e. after this validation runs. So every `pgpu` → `sriovVgpu` / `migBackedVgpu` request with profiles died before hex was called.

```diff
-if isVgpuType(h.gpuCardReq.ResourceType) && len(h.gpuCardReq.Profiles) > 0 {
-    profilesMap, _, err = getNodeVgpuProfilesMap(card.Id)
-    ...
-    totalVramMiB, err = getGpuTotalVramMiB(card.Id)   // 500s on a vfio-bound card
-    if err != nil { return err }
+if isVgpuType(h.gpuCardReq.ResourceType) && len(h.gpuCardReq.Profiles) > 0 {
+    profilesMap, _, err = getNodeVgpuProfilesMap(card.Id)
+    ...
+    if h.gpuCardReq.ResourceType == gpu.ResourceTypeMigBackedVgpu {
+        totalVramMiB = getGpuTotalVramMiB(card.Id)     // *int; nil = unknown
+    }
 }
```

Two things were wrong, and both are fixed:

| | before | after |
| --- | --- | --- |
| who needs the total | gathered for `isVgpuType()` — includes SR-IOV | only `migBackedVgpu`, matching its single use at `gpu.go:500` |
| card not reportable | error → 500 | `nil` → skip the VRAM rule, log a warning |

**Skipping is not a bypass.** `config_gpu.cpp`'s `ValidateVgpuProfiles` (L981-991) enforces the same rule inside `gpu_resource_set` — *after* `gpu_unbind_vfio_pci`, so it reads a card nvidia-smi can actually see — and it **fails closed**, rolling the card back to vfio-pci if it trips. The API-side check was a shadow of that one, running at the only point in the sequence where it could never succeed.

This also matches what the read path already does: `gpu.go:261` degrades a card nvidia-smi cannot see rather than failing it.

### 2. A resource switch gets 180s instead of 30s

Carving a GPU takes **25–60s** on real hardware (release from vfio-pci → 48 VFs or MIG mode + GPU instances → rewrite `/etc/nova/nova.d/gpu.conf` → restart nova). At 30s, `exec.CommandContext` **SIGKILLed hex_config partway through repartitioning a GPU**, and the caller got a `500 signal: killed` for work that then completed anyway. 180s matches the precedent in `firmware.go:84`; the read-only `hex_sdk` queries in that file keep their 30s.

## Verification (cn13, RTX PRO 6000 Blackwell)

Card `GPU-0a5ba9ad-…`, `pgpu` / `idle`. The node's other three cards were `inUse` and untouched; the test card was returned to `pgpu` / `vfio-pci` / `numvfs=0` afterwards.

| operation | elapsed | result |
| --- | --- | --- |
| pgpu → sriovVgpu | 33.4s | **200** (was `500 signal: killed`) |
| sriovVgpu → pgpu | 26.4s | 200 |
| pgpu → migBackedVgpu | 32.3s | 200 |

Landed state: `sriov_numvfs=48`, `driver=nvidia`, VF `0001:c8:00.2` at `current_vgpu_type=1518`, nova alias `nvidia_rtx_pro_6000_blackwell_dc_2b_1518`; for MIG, `mig.mode.current=Enabled` with GI `MIG 1g.24gb+gfx` (profile 47).

Negative cases, using card total 97887 MiB and profile 1546 (2048 MiB, `vmCountLimit=48`) so that `count=48` → 98304 MiB trips **only** the VRAM rule:

| case | result |
| --- | --- |
| unknown profile id, card visible | `400 vgpu profile 99999 not found` |
| over-commit, card **visible** | `409 requested vram 98304 MiB exceeds device total 97887 MiB` |
| over-commit, card **invisible** (pgpu) | `500 exit status 1` — hex caught it (`exceeds the 97887 MiB available`) and rolled the card back to pgpu + vfio-pci |

`go test -race ./internal/apis/v1/handlers/nodes/` green on arm64 and in the x86 jail container. Four cases added: the SR-IOV path asserts nvidia-smi is **not** queried; the MIG path proceeds both when the card is invisible and when nvidia-smi is unavailable; and the limit still bites when nvidia-smi does report the card.

## Deliberately not in scope

The false failure is rarer, not gone. haproxy fronts this API with `timeout server 1m` and nginx's `/cos-api/` takes the 60s `proxy_read_timeout` default, and measured durations (25.3–60.05s) sit right against that ceiling — one run through the proxy took 60.05s and was cut off with no body while the switch completed; the same operation direct to `:8082` took 32.3s and returned 200. Making the response honest needs the endpoint to go asynchronous (202 + polling `Status.IsProcessing`, which the `ReqGpuCollection` record already drives) — a breaking API change that needs UI coordination, so it belongs in its own ticket. Detail in #638.

### DIFF
--- a/internal/apis/v1/handlers/nodes/gpu.go
+++ b/internal/apis/v1/handlers/nodes/gpu.go
@@ -428,13 +428,14 @@ func isSupportedType(supportTypes []gpu.SupportResourceType, t gpu.ResourceType)
 
 // validateGpuCardUpdate checks a resource-type change against the card's real
 // capabilities. profilesMap (keyed by profile Id) and totalVramMiB (from
-// nvidia-smi) are only consulted for vGPU types with profiles. Returns a
-// sentinel-wrapped error the handler maps to a status.
+// nvidia-smi) are only consulted for vGPU types with profiles; a nil
+// totalVramMiB means the card's total is unavailable and the VRAM check is
+// skipped. Returns a sentinel-wrapped error the handler maps to a status.
 func validateGpuCardUpdate(
 	card gpu.GpuFromHex,
 	req gpu.UpdateGpuCardRequest,
 	profilesMap map[uint32]gpu.VgpuProfileFromHex,
-	totalVramMiB int,
+	totalVramMiB *int,
 ) error {
 	if !isSupportedType(card.SupportTypes, req.ResourceType) {
 		return fmt.Errorf("gpu card %s does not support '%s' resource type: %w", card.Id, req.ResourceType, gpu.ErrUnsupportedType)
@@ -496,10 +497,16 @@ func validateGpuCardUpdate(
 	}
 
 	// The VRAM limit only applies to MIG-backed vGPU; SR-IOV profiles are fixed
-	// partitions constrained by the profile-count limit instead.
-	if req.ResourceType == gpu.ResourceTypeMigBackedVgpu && totalVramReqMiB > totalVramMiB {
+	// partitions constrained by the profile-count limit instead. A nil total
+	// means nvidia-smi could not report the card - the normal state of a
+	// still-vfio-bound pgpu - so the check is skipped here and left to
+	// config_gpu.cpp's ValidateVgpuProfiles, which enforces the same rule after
+	// the card is released from vfio-pci and fails closed. See
+	// getGpuTotalVramMiB.
+	if req.ResourceType == gpu.ResourceTypeMigBackedVgpu &&
+		totalVramMiB != nil && totalVramReqMiB > *totalVramMiB {
 		return fmt.Errorf("gpu %s: requested vram %d MiB exceeds device total %d MiB: %w",
-			card.Id, totalVramReqMiB, totalVramMiB, gpu.ErrExceedVramLimit)
+			card.Id, totalVramReqMiB, *totalVramMiB, gpu.ErrExceedVramLimit)
 	}
 
 	return nil
@@ -841,10 +848,10 @@ func (h *helper) updateLocalGpuCard() error {
 		return err
 	}
 
-	// vGPU profile/VRAM limits need the GPU's available profiles (hex) and its
-	// total VRAM (nvidia-smi). Only gather them when profiles are actually supplied.
+	// vGPU profile limits need the GPU's available profiles (hex). Only gather
+	// them when profiles are actually supplied.
 	var profilesMap map[uint32]gpu.VgpuProfileFromHex
-	var totalVramMiB int
+	var totalVramMiB *int
 	if isVgpuType(h.gpuCardReq.ResourceType) && len(h.gpuCardReq.Profiles) > 0 {
 		profilesMap, _, err = getNodeVgpuProfilesMap(card.Id)
 		if err != nil {
@@ -852,10 +859,13 @@ func (h *helper) updateLocalGpuCard() error {
 			return err
 		}
 
-		totalVramMiB, err = getGpuTotalVramMiB(card.Id)
-		if err != nil {
-			log.Errorf("gpu(%s): failed to get total vram for gpu %s: %v", h.reqId, h.gpuId, err)
-			return err
+		// The card's total VRAM is only ever consulted for MIG-backed vGPU;
+		// SR-IOV profiles are fixed partitions bounded by the profile-count
+		// limit instead (see validateGpuCardUpdate). Querying nvidia-smi for
+		// SR-IOV as well made a pgpu -> sriovVgpu switch fail on a value that
+		// request would never have used.
+		if h.gpuCardReq.ResourceType == gpu.ResourceTypeMigBackedVgpu {
+			totalVramMiB = getGpuTotalVramMiB(card.Id)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


raven-pan (approved):
LGTM!

